PR #8, opened 2015-07-04
Url shortcut

kaihendry:
Handy little script you might like the look of

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,13 @@ To use it, start the `clipmenud` daemon, and then call `clipmenu` to launch
 
 [dmenu]: http://tools.suckless.org/dmenu/
 [xsel]: http://www.vergenet.net/~conrad/software/xsel/
+
+## [dwm](http://dwm.suckless.org/) configuration
+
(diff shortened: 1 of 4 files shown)


cdown:
I'm not sure that creating wrapper scripts for each use case is a good idea. Besides, people shouldn't be blindly executing URLs, at least in this case, I think the most sensible thing would be to grab them from the files and allow selection within dmenu as well (although, perhaps that's just how I would want to use such a script).

Either way, I'd rather not bundle this with dmenu. I will merge the sorting patch soon, though :-)

Not a bad concept in general, though.

kaihendry:
Hi Chris, It's hardly blind. You make a explicit selection first. Maybe link it from somewhere at least? Why not get this project on http://git.suckless.org/ ? :) 20h uses it IIUC.

cdown:
I'm not sure I understand how it isn't blind -- dmenu isn't executed, as far as I can tell it grabs the last URL and executes it without any confirmation upon execution.

I don't really have any interest in moving to git.suckless.org, since I don't see any benefit that that would provide.

PR #9, opened 2015-07-07
added dmenu config variable & started experimenting with xdotool

benjcalderon:
I really like your tool!

what I did is just added some dmenu configs to make it look like the current dmenu that I'm using and added a couple of lines to experiment on having this script actually paste what I  select on the program :-)

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# added an extra variable to add dmenu options
+dmenu_config="-nb #252527 -sb #404E5E -fn CamingoCode-9 -h 46"
 dmenu_lines=${CLIPMENU_LINES-8}
 
 declare -A selections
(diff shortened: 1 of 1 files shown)


cdown:
I don't understand the point of `dmenu_config` at all, just pass in your dmenu args to the script, they will work.

Using xdotool to press a certain sequence of keys that you assume map to a "paste" action is a recipe for disaster.

I won't merge anything involving hardcoding dmenu variables, since the whole point is that they are transparently passed to dmenu -- just pass them to the script (see the readme).

I would consider one that has a sensible method to paste directly.

kaihendry:
Maybe instructions how to customise clipmenu from dwm's `config.h`? E.g.

    static const char dmenufont[] = "-*-terminus-medium-r-*-*-18-*-*-*-*-*-*-*";
    static const char *clipcmd[]  = { "clipmenu", "-fn", dmenufont, NULL };

cdown:
I don't see why one should tell people how to edit `config.h` in something that is totally unrelated to `config.h`...

kaihendry:
Just an example to show how clipmenu is run...

cdown:
That's fair enough, although I don't think adding examples for config.h makes sense by itself.

I added some information in the readme at d5a944917.

cdown:
Closing this out for now, but I'm still open to a patch with xdotool if you'd find that valuable to your workflow. :-)

benjcalderon:
I have been working on other projects but I'll take a look at this later. What you commented about hardcoding dmenu variables makes total sense... I really didn't followed your logic until I took a closer look. I'll experiment a bit more with xdotool or other alternative, what I'm trying to accomplish is something similar to what other clipboard managers do, say, when a text is selected is automatically pasted. Thanks anyways!

PR #11, opened 2015-08-05
fix(clipmenu): update xsel usage

kingcody:
Changes:
* `xsel` now requires an input flag `(-a|-f|-i)` see [`XSEL(1x)`](http://manned.org/xsel/843cf846)

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -34,7 +34,7 @@ chosen_line=$(
 
 for selection in clipboard primary; do
     if type -p xsel >/dev/null 2>&1; then
-        xsel --"$selection" < "${selections[$chosen_line]}"
+        xsel -i --"$selection" < "${selections[$chosen_line]}"
     else
         xclip -sel "$selection" < "${selections[$chosen_line]}"
     fi


cdown:
This isn't backwards-compatible, I'm guessing?

Parsing `xsel --version` seems icky, but perhaps it's the best way. Otherwise we can try one and if it fails try the other, but that kinda sucks too.

cdown:
Hm, it seems it is backwards compatible, but I can't repro that this option is "required" using xsel 1.2.0, which appears to be the latest version:

```
% xsel --version
xsel version 1.2.0 by Conrad Parker <conrad@vergenet.net>
% xsel --clipboard <<< hello
% xsel -o --clipboard          
hello
```

Why do you think that `xsel` now requires an input flag? Since what version?

kingcody:
Well after debugging more since you mentioned not being able to reproduce; it would seem that is only reproducible when wrapped in an `i3-msg exec` (or bound to key via i3 config) like:
```bash
➤ xsel -i --clipboard <<< test1
➤ xsel -o --clipboard
test1
➤ i3-msg "exec xsel --clipboard <<< test2"
[{"success":true}]
➤ xsel -o --clipboard
test1
➤ i3-msg "exec xsel -i --clipboard <<< test2"
[{"success":true}]
➤ xsel -o --clipboard
test2
```

I haven't tried to debug i3 any deeper. Also, sorry for the erroneous statement that `xsel` requires an input flag; I should have checked via shell first.

cdown:
That's bizarre. I have no idea why that would happen, but if you can consistently repro that, I don't mind adding `-i`.

kingcody:
Bizarre indeed, and thanks.

side question:
Not sure this would deserve an issue or anything, but would you mind submitting your `PKGBUILD` for this repo to https://aur4.archlinux.org ? The switch occurs on Aug 8, and it'd be nice to continue to update clipmenu via `yaourt` (:

cdown:
On my to do list (for a while now, admittedly...), I'll probably get around to it later today or tomorrow. Thanks for reminding me though :-)

kingcody:
So many todos... (always) Thanks again.

kingcody:
So just to clarify: Yes, the issue is consistently reproducible with the latest(`4.10.3-1`) `i3-wm`, and after several reboots. (have not tried with older versions of i3)

PR #19, opened 2015-09-11
Take over the clipboard to persist it after the owner quits

Streetwalrus:
I often copy things from Neovim, after which I either quit or suspend (^z) to paste in another instance. If I quit, the clipboard contents are lost, if I suspend, pasting blocks because Neovim is suspended and the destination application can't communicate with it. This PR fixes it and should work with any other application.

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -9,8 +9,10 @@ while sleep "${CLIPMENUD_SLEEP:-0.5}"; do
     for selection in clipboard primary; do
         if type -p xsel >/dev/null 2>&1; then
             data=$(xsel --"$selection"; printf x)
+            xsel --"$selection" | xsel -i --"$selection"
         else
             data=$(xclip -o -sel "$selection"; printf x)
+            xclip -o -sel "$selection" | xclip -i -sel "$selection"
         fi
 
         # We add and remove the x so that trailing newlines are not stripped.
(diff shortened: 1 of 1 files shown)


cdown:
The idea is good, but putting this in the code without a comment about what it is supposed to do will likely confuse future developers unfamiliar with the codebase (or with how X clipboards work). Something along the lines of "Take ownership of clipboard, in case the original application is unable to serve the clipboard request" should do.

Thanks for the patch!

Streetwalrus:
Quick note, while using this patch I noticed that some applications don't like PRIMARY being stolen from them, for example urxvt which unhighlights text. I don't really care about PRIMARY anyway due to its volatile nature so I just disabled it in my local copy of clipmenu. I'll add a comment and a safeguard against this tomorrow, and possibly disabling of PRIMARY through an environment variable or a command line switch.

Streetwalrus:
There you go, it should be good now.

cdown:
Cool! Thanks a lot for the patches.

I'll test this for a little while on my local machine, and put it in a branch for now. Then, once we're all satisfied that everything seems okay, I'll merge it into develop. :-)

cdown:
Now in branch `own_clipboard`.

cdown:
I've been running this for a while, and it seems fine. Will merge shortly.

Thanks for the patch!

Streetwalrus:
It's been working fine for me as well, glad I could help. :)

PR #29, opened 2016-05-05
remove duplicate entries whilst keeping time order

ohnonot:
i found that clipmenud creates duplicate entries when i selct them again and again in clipmenu.
to avoid this, i added a simple line (sledge hammer method): `rm -f "$cache_dir"/*-$md5`
thus, reselected exisiting entries are pushed to the top of the menu, and old ones deleted.

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -52,6 +52,7 @@ while sleep "${CLIPMENUD_SLEEP:-0.5}"; do
         last_data[$selection]=$data
         last_filename[$selection]=$filename
 
+        rm -f "$cache_dir"/*-$md5
         printf '%s' "$data" > "$filename"
     done
 done


cdown:
In general I'm not against this, but there are a few things:

1. There should be a way to opt out, since this was not the default behaviour before
2. The shellcheck error should be fixed (the `mkdir` one is from the base build, I'll fix that separately)
3. Why `-f`?

ohnonot:
1. ok, i think i can look into that.
2. i don't follow. what shellcheck error? this is the first time i do a pull request, sorry if i'm missing something obvious.
3. because it fails silently if there's no file with that md5sum yet.

ohnonot:
i experienced a different error, which made me realize that i cloned and forked the develop branch.
after changing to the master branch, that other error (has nothing to do with this pull request) went away.
so i will delete my fork, close this request and start from scratch.
sorry for the noise.

cdown:
for 2., look at the travis ci results above :-)

PR #37, opened 2016-12-09
Perform an insertion after the selection when --insert given

schtibe:
Added an option that enables the tool to insert the selected clipboard entry directly on the focused input.

So far tested this with i3 with a keybinding:

    bindsym $mod+shift+b exec --no-startup-id clipmenu --insert

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -35,6 +35,14 @@ for (( i=${#files[@]}-1; i>=0; i-- )); do
     selections[$first_line]=$file
 done
 
+# provide the option insert, to perform an insertion an the end
+# remove it when it is given, because the rest of the options will
+# be passed to dmenu
+if [[ $1 == '--insert' ]]; then
+    perform_insert=1
+    shift
+fi
+
 # It's okay to hardcode `-l 8` here as a sensible default without checking
 # whether `-l` is also in "$@", because the way that dmenu works allows a later
 # argument to override an earlier one. That is, if the user passes in `-l`, our
@@ -50,3 +58,8 @@ for selection in clipboard primary; do
         xclip -sel "$selection" < "${selections[$chosen_line]}"
     fi
 done
+
+# perform the insertion
+if [[ $perform_insert == 1 ]]; then
+    xdotool key shift+Insert
+fi


cdown:
Thanks for contributing. Whether this makes sense or not depends on your window manager's behaviour when reverting focus to the previous client. It probably *should* work in the majority of cases, so I'm not too worried about this, but I do wonder if we will end up with bugs filed that are due to WM behaviour.

As for the implementation itself, I have a few questions. What's the benefit over `clipmenu && xdotool key shift+Insert`? I'm reluctant to add code that has no tangible benefit to being integrated, as opposed to just being prepended or appended to existing clipmenu workflows by normal shell flow control.

If we agree to merge this patch after discussion, here are a few things I think are worth revising:

1. I'm unlikely to force people to install xdotool by default -- it will probably go in `optdepends` on Arch Linux, for example. As such, we should detect the case that people do not have xdotool installed and print a warning on stderr that they need to install it for this feature.
2. I think having to put `--insert` first for this to work is a bit non-intuitive. I'd prefer a solution similar to the one I've used in `mpdmenu`: https://github.com/cdown/mpdmenu/blob/master/mpdmenu#L36-L48
3. `(( perform_insert ))`, not `[[ $perform_insert == 1 ]].` `((` generally acts in a more reasonable way when doing arithmetic comparison.
4. Typically, `Shift+Insert` only pastes `PRIMARY` (although some applications do use it for `CLIPBOARD`). It seems unexpected that this feature can only support one clipboard type. It would be good to have it generalised so that you can pass the keys you want pressed.

Thanks! :-)

schtibe:
Thanks for your elaborate response!

> What's the benefit over clipmenu && xdotool key shift+Insert

Good question. Implementing this turned out to be way easier than anticipated.  One benefit I see is that it would be available for all users that don't come up with the idea by themselves. But this could of course be solved by adding it to the README as well!
Something else: are all WMs capable of concatting commands when making a key binding? (It could still be done via intermediat script though)

A real benefit of having it in the code would maybe arise when we consider other implementations than Shift+Insert, as you mentionned in 4.

FYI I was told that Shift+Insert was the most common inserting shortcut found in applications. I also considered sending the selection as text, but I wonder if this leads to encoding errors and the like.

Anyway I absolutely concur with your considerations and wouldn't mind you rejecting this patch. I like the Unix philosophy after all ;)

cdown:
I'll leave this for now, but thanks for the contribution and your understanding :-)

PR #44, opened 2017-03-08
Add `--logfile /dev/null` to the xsel command

drake-on-mandrake:
@kesha-on-disc forgot part of her line in #36.


cdown:
Seems legit, thanks!

PR #46, opened 2017-03-26
Ability to sync clipboard

its-me:


cdown:
IMO this logic is wrong. Synchronisation should be by time, not only `PRIMARY` to `CLIPBOARD`.

cdown:
Synchronisation should be by time, not only `PRIMARY` to `CLIPBOARD`.

I'm also not entirely sure that synchronising clipboards is something clipmenu should be the business of doing -- any reason you think it should be here (instead of in, say, autocutsel)?

PR #68, opened 2018-03-09
Allow use as a rofi script

RX14:
Rofi has a mode called "script mode" where rofi itself launches the script. The script is expected to echo it's output to stdout then exit. When a selection is made, the script is called again, with the selected text as the first argument. This mode of operation allows rofi to show different scripts in multiple tabs (i.e. I have one rofi instance which is an application launcher AND a clipboard manager)

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -35,11 +35,11 @@ list_clips() {
 }
 
 if [[ "$CM_LAUNCHER" == rofi-script ]]; then
-    if [[ $# -eq 0 ]]; then
+    if ! (( $# )); then
         list_clips
         exit
     else
-        chosen_line="$1"
+        chosen_line="${@: -1}"
     fi
 else
     # It's okay to hardcode `-l 8` here as a sensible default without checking
--- a/clipmenu
+++ b/clipmenu
@@ -39,6 +39,8 @@ if [[ "$CM_LAUNCHER" == rofi-script ]]; then
         list_clips
         exit
     else
+        # https://github.com/koalaman/shellcheck/issues/1141
+        # shellcheck disable=SC2124
         chosen_line="${@: -1}"
     fi
 else
--- a/clipmenu
+++ b/clipmenu
@@ -30,14 +30,26 @@ if [[ "$CM_LAUNCHER" == rofi ]]; then
     set -- -dmenu "$@"
 fi
 
-# It's okay to hardcode `-l 8` here as a sensible default without checking
-# whether `-l` is also in "$@", because the way that dmenu works allows a later
-# argument to override an earlier one. That is, if the user passes in `-l`, our
-# one will be ignored.
-chosen_line=$(
-    cat "$cache_file_prefix"_* /dev/null | LC_ALL=C sort -rnk 1 |
-        cut -d' ' -f2- | awk '!seen[$0]++' | "$CM_LAUNCHER" -l 8 "$@"
-)
+list_clips() {
+    cat "$cache_file_prefix"_* /dev/null | LC_ALL=C sort -rnk 1 | cut -d' ' -f2- | awk '!seen[$0]++'
+}
+
+if [[ "$CM_LAUNCHER" == rofi-script ]]; then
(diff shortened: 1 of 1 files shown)


clipmenu
```diff
@@ -30,14 +30,26 @@ if [[ "$CM_LAUNCHER" == rofi ]]; then
...
-chosen_line=$(
-    cat "$cache_file_prefix"_* /dev/null | LC_ALL=C sort -rnk 1 |
-        cut -d' ' -f2- | awk '!seen[$0]++' | "$CM_LAUNCHER" -l 8 "$@"
-)
+list_clips() {
+    cat "$cache_file_prefix"_* /dev/null | LC_ALL=C sort -rnk 1 | cut -d' ' -f2- | awk '!seen[$0]++'
+}
+
+if [[ "$CM_LAUNCHER" == rofi-script ]]; then
+    if [[ -z "$@" ]]; then
```
cdown: if `! (( $# ))`

clipmenu
```diff
@@ -30,14 +30,26 @@ if [[ "$CM_LAUNCHER" == rofi ]]; then
...
+        list_clips
+        exit
+    else
+        chosen_line="$@"
```
cdown: If this is how rofi inputs the new selection, it's unlikely to work very well with our policy to preserve whitespace accurately. Or maybe it's really just `$1`? I'll have a look myself this weekend.

RX14: It's really just `$1`, I can change this.

clipmenu
RX14: How about `[[ $# -eq 0 ]]`?

RX14:
Pushed an update commit

clipmenu
```diff
@@ -35,11 +35,11 @@ list_clips() {
 }
 
 if [[ "$CM_LAUNCHER" == rofi-script ]]; then
-    if [[ -z "$@" ]]; then
+    if [[ $# -eq 0 ]]; then
```
cdown: as mentioned in last review, use `((` for arithmetic comparison in bash, not `[[` :-)

clipmenu
```diff
@@ -35,11 +35,11 @@ list_clips() {
...
         list_clips
         exit
     else
-        chosen_line="$@"
+        chosen_line="$1"
```
cdown: Based on your comment, maybe this should be `chosen_line=${@: -1}`?

clipmenu
RX14: Sure, but i'd like to understand why to improve my own scripts in the futue

cdown: Mostly stylistic reasons. It allows for easy checking of existence/non-existence and we already use this all over the place, for example in clipmenud you can see a lot of `(( some_flag ))`, which is more terse, and easier to skim through.

In general I try to only ever use `[[` for non-arithmetic comparison, and always use `((` for arithmetic since it usually expresses whatever is desired in a more compact and readable way. In this case it's somewhat negligible, but I'd prefer to keep the style we already use :-)

RX14: OK, makes a lot of sense.

RX14:
Fixed.

cdown:
Looks great to me! Can you give some instructions to me on how to test it? I found this document, but not much else: https://github.com/DaveDavenport/rofi/wiki/rfc-script-mode

cdown:
Meh, that shellcheck advice is wrong. I've filed https://github.com/koalaman/shellcheck/issues/1141, but for now we also need to silence it. You can add this just above to turn off the warning for the next line:

```sh
# https://github.com/koalaman/shellcheck/issues/1141
# shellcheck disable=SC2124
```

cdown:
Thanks! Still would be ace to know how to use this so we can put something in the README, but looks more than ready to merge.

RX14:
I personally use it like
```
rofi -modi 'drun,clipmenu:env CM_LAUNCHER=rofi-script ./clipmenu,window,run' -show drun
```

Issue #73, opened 2018-04-03
Clipmenu stops pasing selection


stef204:
I have used clipmenu only for a couple of days but have run into one particular issue:

At some point, it will stop pasting what I select and only paste the top item.  I usually try many times, and try to use other lines, etc., and it seems erratic, pasting some selections but not others, etc.

I see this: `: "${CM_MAX_CLIPS=1000}"` (pasted here with clipmenu, mind you!)

In what unit is the 1000? 1000 lines, characters, something else?

This could be happening only when clipboard is full but not sure yet.

cdown:
Maybe this is #71 or #70. Thanks for the clear description.

1000 is the number of clips, ie. we only keep 1000 clipboard entries in cache. I'll do some testing in a bit.

stef204:
Happy to test and help troubleshoot, just let me know what is needed.

cdown:
Thanks, I've not been able to repro thus far. The best thing to do would be to show `bash -x "$(command -v clipmenu)"`, select a clip, and give me an example where it selected the wrong thing.

stef204:
If I run above command, it brings up rofi/clipmenu. I then either use a clip by pressing enter. and It  returns to command line, or presse `Escape`, exiting in both cases.

IOW, you want to me to stop using `clipmenu` and substitute that for `bash -x "$(command -v clipmenu)"` every single time until the problem occurs. (Best would be I modify the KB shortcut I set up.)  Is that correct?

cdown:
Yeah, that would be best if it's not reproducible every time.

You might want to redirect the output to a file so that you can get it afterwards.

stef204:
You're going to have to help me out; having trouble redirecting to a file (sorry, I cannot take the time to `RTFM bash` at the moment due to work.)

I am using a script (for which I have created a KB shortcut):

    #!/bin/sh

    bash -x "$(command -v clipmenu)" >> $HOME/cdb.log
    exit

It does not redirect to the file which I created using `touch cdb.log`

I tried a number of pipes, etc., but I am obviously missing the info on this very basic point.

I tried `2>&1 > cdb.log` but no joy there either.

How do I amend the script to capture output in a way it appends and does not overwrite?

stef204:
Also, I _think_ I may have found something, clipmenu _may_ have trouble with symbols or path names, etc.  That is how the "bug" showed up first.  And it is now back again:

[Here](https://paste.xinu.at/08sL/) is the log.

[...]

PR #76, opened 2018-04-19
Fix wrong file deleted after partial selection

### LINKED ISSUE #75: Selection not in cache after partial
When copying subsequent partial selection, the selection just before is removed from cache.
I found something in `clipmenud` that might be a bug, here is a possible fix:

```diff
diff --git a/clipmenud b/clipmenud
index 3c4d27b252..4eba9acfd3 100755
--- a/clipmenud
+++ b/clipmenud
@@ -181,9 +181,6 @@ while true; do
             rm -- "${last_filename[$selection]}"
         fi
 
-        last_data[$selection]=$data
-        last_filename[$selection]=$filename
-
         first_line=$(get_first_line "$data")
 
         debug "New clipboard entry on $selection selection: \"$first_line\""
@@ -191,6 +188,9 @@ while true; do
         cache_file_output="$(date +%s%N) $first_line"
         filename="$cache_dir/$(cksum <<< "$first_line")"
 
+        last_data[$selection]=$data
+        last_filename[$selection]=$filename
+
         # Recover without restart if we deleted the entire clip dir.
         # It's ok that this only applies to the final directory.
         # shellcheck disable=SC2174
```

Gravemind:
It seems it was deleting the file before the partial selection (instead of the last one, the partial selection), because `last_filename`  was saved before filename was actually ready (so was
actually saving the filename before the last one).

Fixes #75.

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -181,15 +181,15 @@ while true; do
             rm -- "${last_filename[$selection]}"
         fi
 
-        last_data[$selection]=$data
-        last_filename[$selection]=$filename
-
         first_line=$(get_first_line "$data")
 
         debug "New clipboard entry on $selection selection: \"$first_line\""
 
         cache_file_output="$(date +%s%N) $first_line"
         filename="$cache_dir/$(cksum <<< "$first_line")"
+
+        last_data[$selection]=$data
+        last_filename[$selection]=$filename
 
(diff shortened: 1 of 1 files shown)


cdown:
Oh jeez. Thanks. This is probably another part of #73 as well.

PR #83, opened 2018-08-11
 clipmenu: automatic selection of CM_LAUNCHER

moviuro:


clipmenu
```diff
@@ -1,7 +1,17 @@
 #!/bin/bash
 
-: "${CM_LAUNCHER=dmenu}"
 : "${CM_DIR="${XDG_RUNTIME_DIR-"${TMPDIR-/tmp}"}"}"
+# Pick dmenu or rofi, depending on what's available on the system
+if [[ -z "$CM_LAUNCHER" ]]; then
```
cdown: if ! [[ -v CM_LAUNCHER ]]; then

clipmenu
```diff
@@ -1,7 +1,17 @@
...
+  if command -v dmenu >/dev/null 2>&1; then
+    CM_LAUNCHER=dmenu
```
cdown: indentation in this patch is wrong, it should be 4 spaces per indent.

clipmenu
```diff
@@ -1,7 +1,17 @@
...
+  elif command -v rofi >/dev/null 2>&1; then
+    CM_LAUNCHER=rofi
+  else
+    echo "Neither dmenu(1) nor rofi(1) were found." >&2
+    exit 5
```
cdown: thanks for keeping distinct error codes :-)

clipmenu
```diff
@@ -1,7 +1,17 @@
 #!/bin/bash
 
-: "${CM_LAUNCHER=dmenu}"
 : "${CM_DIR="${XDG_RUNTIME_DIR-"${TMPDIR-/tmp}"}"}"
+# Pick dmenu or rofi, depending on what's available on the system
```
cdown: this comment just states the obvious, you can remove it

cdown:
Hi! This looks fine, only some minor things before merge.

clipmenu
moviuro: Yeah, my bad, I wrote that in a hurry... I'll fix it tomorrow!

clipmenu
moviuro: Why? This is further from POSIX syntax I'm used to. Is it faster?...

clipmenu
moviuro: Yeah, my bad, I wrote that in a hurry... I'll fix it tomorrow!

clipmenu
cdown: We already use lots of non-POSIX features, so trying to keep close compatibility isn't a goal. `-v` has clearer intent than `-z`. This isn't a massive deal but I'd rather stick to options which semantically represent the intent more clearly.

cdown:
Hey! This still looks fine to me -- do you intend to update it for 6.0.0? :-)

cdown:
Closing for now without prejudice -- happy to reopen if you intend to send an updated patch :-)

Issue #86, opened 2018-08-25
Disable recording for certain windows (e.g. password managers)


Sohalt:
It would be cool if I could configure clipmenu to not store text copied from windows with a certain title, to avoid storing passwords copied from a password manager.
This should be fairly straightforward to implement using `xdotool getactivewindow getwindowname 
` to get the title of the currently active window and checking it against a regex before storing the clipboard.

CopyQ supports this:
https://copyq.readthedocs.io/en/latest/faq.html#how-to-omit-storing-text-copied-from-specific-windows-like-a-password-manager

cdown:
I'd accept a patch doing this. :-)

Sohalt:
I'll give it a go.

Sohalt:
This is implemented in #89. Can you please have a look?

PR #89, opened 2018-10-09
Add support to ignore the clipboard in certain windows

Sohalt:
The clipboard does not get recorded when the title of the currenlty active
window matches the regular expression in CM_IGNORE_WINDOW. This allows copying
passwords from a password manager without the passwords ending up in clipmenu.

The matching is not 100% exact however, as there is a race condition bewteen the
time the clipboard is populated, clipmenu queries the clipboard, and the active
window gets queried. This race condition can be especially problematic when
using polling with large intervals instead of clipnotify.

This implements #86.

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -16,6 +16,7 @@ cache_file_prefix=$cache_dir/line_cache
 lock_file=$cache_dir/lock
 lock_timeout=2
 has_clipnotify=0
+has_xdotool=0
 
 # This comes from the environment, so we rely on word splitting.
 # shellcheck disable=SC2206
@@ -105,6 +106,12 @@ command -v clipnotify >/dev/null 2>&1 && has_clipnotify=1
 if ! (( has_clipnotify )); then
     echo "WARN: Consider installing clipnotify for better performance." >&2
     echo "WARN: See https://github.com/cdown/clipnotify." >&2
+fi
+
+command -v xdotool >/dev/null 2>&1 && has_xdotool=1
+
+if [[ "$CM_IGNORE_WINDOW" && ! (( has_xdotool )) ]]; then
+    echo "WARN: CM_IGNORE_WINDOW does not work without xdotool, which is not installed" >&2
 fi
 
 exec {lock_fd}> "$lock_file"
--- a/clipmenud
+++ b/clipmenud
@@ -129,10 +129,12 @@ while true; do
         fi
     fi
 
-    windowname="$(xdotool getactivewindow getwindowname)"
-    if [[ -n "${CM_IGNORE_WINDOW+x}" && "$windowname" =~ $CM_IGNORE_WINDOW ]]; then
-        debug "ignoring clipboard because windowname \"$windowname\" matches \"${CM_IGNORE_WINDOW}\""
-        continue
+    if [[ "${CM_IGNORE_WINDOW}" && (( has_xdotool ))]]; then
+        windowname="$(xdotool getactivewindow getwindowname)"
+        if [[ "$windowname" =~ $CM_IGNORE_WINDOW ]]; then
+            debug "ignoring clipboard because windowname \"$windowname\" matches \"${CM_IGNORE_WINDOW}\""
+            continue
+        fi
     fi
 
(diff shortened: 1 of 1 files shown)


Sohalt:
Maybe the dependency on xdotool should be documented somewhere in the README or help text, as well as the race condition.
And maybe the implemented functionality should only work when clipnotify is installed, because without, the possibility for sensitive data ending up in the clipboard is too high and the feature working somewhat might give the user a false sense of security.

clipmenud
```diff
@@ -121,6 +122,12 @@ while true; do
         fi
     fi
 
+    windowname="$(xdotool getactivewindow getwindowname)"
+    if [[ -n "${CM_IGNORE_WINDOW+x}" && "$windowname" =~ $CM_IGNORE_WINDOW ]]; then
```
cdown: You don't need `+x` or `-n` here

clipmenud
```diff
@@ -121,6 +122,12 @@ while true; do
         fi
     fi
 
+    windowname="$(xdotool getactivewindow getwindowname)"
```
cdown: This should only be executed if we actually have anything to check

cdown:
Looks generally sane, just a few things to change.

clipmenud
Sohalt: I wanted to distinguish between `CM_IGNORE_WINDOW` being unset and `CM_IGNORE_WINDOW` being set to the empty string, so I used the approach from https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash (I thought the empty string would match the empty string and not be treated as a wildcard, so it would be helpful to include). I'll update the code

clipmenud
```diff
@@ -106,6 +108,12 @@ if ! (( has_clipnotify )); then
     echo "WARN: See https://github.com/cdown/clipnotify." >&2
 fi
 
+command -v xdotool >/dev/null 2>&1 && has_xdotool=1
+
+if [[ "$CM_IGNORE_WINDOW" && ! (( has_xdotool )) ]]; then
```
cdown: You can't use `((` inside `[[` like that, it will always be true :-) You need to use `((`in keyword-eligible scope

clipmenud
```diff
@@ -121,6 +129,14 @@ while true; do
         fi
     fi
 
+    if [[ "${CM_IGNORE_WINDOW}" && (( has_xdotool ))]]; then
```
cdown: ditto

clipmenud
```diff
@@ -86,6 +87,7 @@ Environment variables:
 - $CM_ONESHOT: run once immediately, do not loop (default: 0)
 - $CM_OWN_CLIPBOARD: take ownership of the clipboard (default: 1)
 - $CM_SELECTIONS: space separated list of the selections to manage (default: "clipboard primary")
+- $CM_IGNORE_WINDOW: disable recording the clipboard in windows where the windowname matches the given regex (e.g. a passwordmanager), do not ignore any windows if unset (default: unset)
```
cdown: `s/passwordmanager/password manager/` :)

Sohalt:
```suggestion
- $CM_IGNORE_WINDOW: disable recording the clipboard in windows where the windowname matches the given regex (e.g. a password manager), do not ignore any windows if unset (default: unset)
```

clipmenud
cdown: You want `if [[ $CM_IGNORE_WINDOW ]] && (( has_xdotool )); then` :-)

Sohalt: Thanks!
```suggestion
if [[ $CM_IGNORE_WINDOW ]] && ! (( has_xdotool )); then
```

clipmenud
Sohalt:
```suggestion
    if [[ $CM_IGNORE_WINDOW ]] && (( has_xdotool )); then
```

cdown:
Thanks! Just waiting for CI.

Sohalt:
Do you want me to squash all the commits into one?

cdown:
No worries, I can squash them on merge.

cdown:
Thanks! Merged and will go into next release.

PR #91, opened 2018-10-27
clipdel: Support getting pattern from standard input

ferki:
I'm using `clipmenu` together with `pass` (and `passmenu` :)), and
wanted to make sure that `pass` would remove passwords both from the
clipboard and the clipboard history.

I had to slightly patch `pass` for that locally, where I also figured it
would be best to simply pipe the password string into `clipdel`, instead
of passing it as an argument.

This patch makes `clipdel` look for a pattern coming from standard input
if there's no pattern passed as an argument.

Please review, and let me know if you need me to improve it, or to open
an issue instead for further discussions.

### DIFF
--- a/clipdel
+++ b/clipdel
@@ -40,8 +40,6 @@ if ! [[ -f $cache_file ]]; then
     exit 0  # Well, this is a kind of success...
 fi
 
-# https://github.com/koalaman/shellcheck/issues/1141
-# shellcheck disable=SC2124
 raw_pattern=$1
 esc_pattern=${raw_pattern//\#/'\#'}
 
--- a/clipdel
+++ b/clipdel
@@ -19,7 +19,8 @@ lock_timeout=2
 if [[ $1 == --help ]] || [[ $1 == -h ]]; then
     cat << 'EOF'
 clipdel deletes clipmenu entries matching a regex. By default, just lists what
-it would delete, pass -d to do it for real.
+it would delete, pass -d to do it for real. If no pattern is passed as an argument,
+it will try to read one from standard input.
 
 ".*" is special, it will just nuke the entire data directory, including the
 line caches and all other state.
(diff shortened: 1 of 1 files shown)


ferki:
Thanks for your patience here!

I'd like to update this PR, and most of the steps are clear to me (rebase on top of current master, and replace `cat` usage with `IFS= read -r`).

I think I need some clarifications on the feedback about checking for a connected terminal, though:

> 1. User experience is not great if invoked incorrectly now (ie. you don't get the "no pattern provided, see --help" message). This can be mitigated by checking if connected to a terminal with `[[ -t 1 ]]`.

Where should this `[[ -t 1 ]]` check go? Into the `if` that decides about printing the error message, or into a new `if` that decides how to obtain a value for `raw_pattern`? Perhaps both, or even somewhere else?

cdown:
Sorry this went stale -- this is still okay, though. Probably the terminal check can just be around `raw_pattern` assignment itself.

ferki:
No worries, thanks for the feedback!

I've rebased my changes on top of current master, and made a more explicit check for the various sources of possible input. Let me know how it fits the codebase now!

clipdel
```diff
@@ -40,10 +41,13 @@ if ! [[ -f $cache_file ]]; then
...
-# shellcheck disable=SC2124
-raw_pattern=$1
-esc_pattern=${raw_pattern//\#/'\#'}
+if [[ -n $1 ]]; then
+    raw_pattern=$1
+elif ! [[ -t 0 ]]; then
+    IFS= read -r raw_pattern
+fi
+
+esc_pattern=${raw_pattern/\#/'\#'}
```
cdown, line 50: Hmm, why only one? We want to update all of them in the string, no?

ferki: Well spotted, thanks! It's an unintended change. At the time I forked, clipdel was still only escaping the first one, and I missed the single-character change in this line during rebase.

Fixed.

cdown:
Thanks! Looks good to me.

PR #94, opened 2018-11-12
Fix clipdel cutting timestamps from file cache

Gravemind:
The actual bug I encountered was that clipdel apparently re-ordered the clipmenu entries (seen when writing [my own clipmenu-del](https://github.com/Gravemind/dotfiles/blob/master/bin/clipmenu-del)).

Looking into it, I found clipdel `cut`ed the timestamp column from the line cache file!

So, I tried to fix this, and keep the pattern retro-compatible:
- still using the pattern in `sed`
- not using `sed -E`
- not break `clipdel -d '^exact_match$'` (the pattern is matched only against the text, without the timestamp).

Turns out sed is quite powerful, I discovered more [sed commands](https://www.gnu.org/software/sed/manual/html_node/Other-Commands.html#Other-Commands) that did the trick. But **there could be a better solution I didn't see**.

sed command `h` and `g` seems standard (at least [non-gnu compatible](http://netbsd.gw.com/cgi-bin/man-cgi?sed++NetBSD-current)), but **I did not actually test it with a non-gnu sed** (on bsd ?). 

Also, I thought of using `sed -i` instead of for+mktemp, but I know there might be compatibility issues with non-gnu-sed, so I dropped it...

### DIFF
--- a/clipdel
+++ b/clipdel
@@ -47,6 +47,10 @@ fi
 raw_pattern=$1
 esc_pattern=${raw_pattern//\#/'\#'}
 
+# We use 2 separate sed commands so "esc_pattern" matches only the 'clip' text
+# without the timestamp (e.g. $> clipdel '^delete_exact_match$')
+sed_common_command="s#^[0-9]\+ ##;\\#${esc_pattern}#"
+
 if ! [[ $raw_pattern ]]; then
     printf '%s\n' 'No pattern provided, see --help' >&2
     exit 2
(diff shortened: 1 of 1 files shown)


Gravemind:
> doing `LC_ALL=C sed "\\#^[0-9]\+ ${esc_pattern}#d"` do?

I thought of that, but then `clipdel -d '^exact_match$'` wouldn't work anymore (breaks the regexp: `^[0-9]+ ^exact_match$`...).

Gravemind:
(Not sure if you were talking about that, but I just re-pushed without a last minute mistake I made in the regex: `[0-9]+` instead of `[0-9]\+` in `s#^[0-9]\+ ##;\\#${esc_pattern}#`)

cdown:
Hmm, you're right. Let me think a bit about this and get back to you.

cdown:
Sorry, forgot all about this. Clearly I thought too much, so merged. Thanks for your contribution, sorry for the delay.

PR #96, opened 2018-12-01
Quit if clipnotify fails

ggVGc:
I think this is more useful behaviour, so possibly better for others too.


cdown:
Hi! Why? What's your justification?

cdown:
Closing due to lack of response.

PR #108, opened 2019-05-30
clipmenu: print selection to stdout

jb55:
This allows you to use clipmenu in desktop scripts. For example you
could pipe the output of your narrowed selection to another command.

Signed-off-by: William Casarin <jb55@jb55.com>

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -23,6 +23,7 @@ Environment variables:
 - $CM_DIR: specify the base directory to store the cache dir in (default: $XDG_RUNTIME_DIR, $TMPDIR, or /tmp)
 - $CM_HISTLENGTH: specify the number of lines to show in dmenu/rofi (default: 8)
 - $CM_LAUNCHER: specify a dmenu-compatible launcher (default: dmenu)
+- $CM_OUTPUT_CLIP: if set, output clip selection to stdout
 EOF
     exit 0
 fi
@@ -52,3 +53,7 @@ file=$cache_dir/$(cksum <<< "$chosen_line")
 for selection in clipboard primary; do
     xsel --logfile /dev/null -i --"$selection" < "$file"
 done
+
+if (( CM_OUTPUT_CLIP )); then
+    cat "$file"
+fi


cdown:
This is fine, but we need to gate it, and we really don't have a good way of doing that right now.

Let me fix up the current options to get replaced by getopt. Then we can back this with `-p` or something. :-)

jb55:
ok, also it might be better to just cat the file and then printf'ing a newline instead of buffering the whole file into memory here. probably not that big of deal, but for large selections it might be slow/use memory.

jb55:
I pushed an update, no need to do anything complicated with getopt (yet). what do you think?

jb55:
ah I just realized -p is a dmenu option, so this will break existing setups

jb55:
pushed a v3 that replaces the `-p,--print` option with `--clipmenu-print` which shouldn't overlap with $CM_LAUNCHER options

cdown:
I'd rather properly support getopt as a separate piece of work, first. I'll take a look at that this week. Then we can just rebase and piggyback this PR off of it :-)

cdown:
In the end we didn't use getopt, because I don't want to end up with mpdmenu-style arg mangling.

If this is gated behind something like `CM_OUTPUT_CLIP`, we should be good to go :-)

jb55:
done

cdown:
Looks great! Let's make it accept 1 or 0 as with the other settings, though, by making the check `(( CM_OUTPUT_CLIP ))` :-)

jb55:
neat trick, will be using that in my scripts now. pushed an updated

jb55:
tests are failing but I'm not sure why

cdown:
`(( CM_OUTPUT_CLIP ))` rather than `(( "$CM_OUTPUT_CLIP" ))` -- there's no need to do that inside arithmetic expressions :-)

It fails because your command returns 1, which is the final command executed, so the script also returns 1.

You can also drop the `exec` -- this is the last thing we run anyway, so there's not much to be gained, especially if refactoring happens later.

jb55:
thanks, updated

jb55:
ok that should fix it...

clipmenu
```diff
@@ -52,3 +53,7 @@ file=$cache_dir/$(cksum <<< "$chosen_line")
 for selection in clipboard primary; do
     xsel --logfile /dev/null -i --"$selection" < "$file"
 done
+
+if (( CM_OUTPUT_CLIP )); then
+  cat "$file"
```
cdown: please match the indentation for the rest of the file :)

jb55:
> please match the indentation for the rest of the file :)

lol fair. sorry for the patch review DoS attack

PR #112, opened 2019-08-23
client: add debug env var to print selected text

mie00:


cdown:
I'm not sure I understand the purpose of this. Could you please explain the rationale? How is this useful for debugging, as opposed to just doing it yourself?

PR #120, opened 2020-03-12
Improve manual install experience

markstos:
- Add Makefile to install files
   - It even checks for required binaries
 - Document manual install steps
 - Document how to start and enable via systemd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+# `dmenu` is not a hard dependency, but you need it unless
+# you plan to set CM_LAUNCHER to another value like `rofi`
+REQUIRED_BINS := xsel clipnotify
+$(foreach bin,$(REQUIRED_BINS),\
+    $(if $(shell command -v $(bin) 2> /dev/null),$(info Found `$(bin)`),$(error Missing Dep. Please install `$(bin)`)))
+
+.PHONY: install
+
+install:
+	install -D -m755 clipmenu /usr/bin/clipmenu
+	install -D -m755 clipmenud /usr/bin/clipmenud
+	install -D -m755 clipdel /usr/bin/clipdel
+	install -D -m644 init/clipmenud.service /usr/lib/systemd/user/clipmenud.service
--- a/README.md
+++ b/README.md
@@ -12,11 +12,13 @@ clipmenu is a simple clipboard manager using [dmenu][] (or [rofi][] with
 Start `clipmenud`, then run `clipmenu` to select something to put on the
 clipboard.
 
-A systemd user service for starting clipmenud is included at
-[init/clipmenud.service](https://github.com/cdown/clipmenu/blob/develop/init/clipmenud.service).
-You can then start clipmenud like this:
+A systemd user service for starting clipmenud is installed as part of the project.
 
-    systemctl --user start clipmenud
+If you want to start `clipmenud` immediately via systemd and have it be started when you login, run:
+
+    systemctl --user enable --now clipmenud
+
+You may wish to bind a shortcut in your window manager to launch `clipmenu`.
 
 All args passed to clipmenu are transparently dispatched to dmenu. That is, if
 you usually call dmenu with args to set colours and other properties, you can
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -23,13 +25,26 @@ invoke clipmenu in exactly the same way to get the same effect, like so:
...
+
+If your distribution doesn't provide a package, you can run the following
+the commands to install this.  (Or better yet, create package for your distribution!).
+You'll first need to install `xsel`, `dmenu` and `clipnotify`.
+
+     git clone https://github.com/cdown/clipmenu.git
+     cd clipmenu
+     sudo make install
+
+If you want to start `clipmenud` immediately via systemd and have it be started whenm you login, run:
```
cdown: s/whenm/when/

Makefile
```diff
@@ -0,0 +1,13 @@
+
+
+REQUIRED_BINS := xsel dmenu clipnotify
```
cdown: dmenu isn't required if you're going to use a different `CM_LAUNCHER`.

README.md
```diff
@@ -14,6 +14,8 @@ clipboard.
 
 A systemd user service for starting clipmenud is included at
 [init/clipmenud.service](https://github.com/cdown/clipmenu/blob/develop/init/clipmenud.service).
+It's installed as part of the manual install process below.
```
cdown: I'd rather not recommend manual installation upfront, people should be using packages where possible.

README.md
```diff
@@ -23,13 +25,26 @@ invoke clipmenu in exactly the same way to get the same effect, like so:
 
 You can remove clips with the `clipdel` utility, see `clipdel --help`.
 
-# Installation
+# Installation - Packages
```
cdown: I'd rather not differentiate this, this is the canonical way to install.

README.md
```diff
@@ -23,13 +25,26 @@ invoke clipmenu in exactly the same way to get the same effect, like so:
...
 Several distributions, including Arch and Nix, provide clipmenu as an official
 package called `clipmenu`.
 
-If your distribution doesn't provide a package, you can run the scripts
-standalone (or better yet, package them!).
+# Installation - Manual
+
+If your distribution doesn't provide a package, you can run the following
+the commands to install this.  (Or better yet, create package for your distribution!).
+You'll first need to install `xsel`, `dmenu` and `clipnotify`.
```
cdown: ditto on dmenu

README.md
markstos: Ubuntu is the most popular desktop Linux distribution and there is no package for it. I was trying to provide easy to find instructions that would help people who are using the most popular Linux distro. How you do recommend documenting the install process for Ubuntu users?

Makefile
markstos: `dmenu` is a hard dependency on Arch Linux. I was modeling what's done in the PKGBUILD file: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/clipmenu

By depending on `dmenu`, you have a better chance of success for a very common case.

Like a lot of `rofi` users, I symlink `dmenu` to rofi so that every `dmenu` call is redirected. Even a number of `rofi` users will have `dmenu` in their path. That's the only way to handle some other `dmenu` utilities.

README.md
markstos: The way the `systemd` file is linked to directly here it's not clear that it's included in the packages (It is for Arch) or would otherwise be included during install. It seems like the suggestion is that the file might need to be copy/pasted or manually installed. 

I don't think people need to see the contents of the `systemd` file any more than they need to see the content of the `bash` scripts. Why not just remove the link and say:

> A systemd user service for starting clipmenud is installed as part of the project.

And just leave it that.

markstos: On the other hand, all users who want to use the `systemd` integration need to actually enable and start the systemd service. I included that in the "manual install" section put perhaps it should be moved up to the "Usage" section, since it applies to those using packages as well. At least Arch doesn't actually `start` or `enable` the service for you.

markstos:
I've pushed an amend commit which addresses the feedback above.

 * The "Manual Install" section has now been minimized, with some more detailed moved to "Usage" that apply to all cases.
 * `dmenu` is no longer considered a dependency for manual installation.

README.md
cdown: Different distributions have different ideas about what the right way to do things is. Arch just happens to (correctly...) have the idea that one shouldn't enable or start things up front. I'm pretty sure Debian and Ubuntu at least historically have taken the opposite perspective, so it's more of a distribution question than a package one.

> I don't think people need to see the contents of the systemd file any more than they need to see the content of the bash scripts.

You're going to probably end up looking at it if you want to customise any of the environment. I mean sure, you can use a dropin, but you're going to go and look at what's there already to work out what to do.

Makefile
```diff
@@ -0,0 +1,15 @@
+
```
cdown: did you mean to leave these blank?

markstos:
I've pushed a refinement to remove the blank lines.

cdown:
This is fine, just needs rebase for 6.0.0. Thanks!

markstos:
Rebased.

cdown:
Thanks!

PR #121, opened 2020-03-13
document options for clipmenu

markstos:
For privacy considerations, it's important to know about CM_MAX_CLIPS
and it's default to storing to 1000 clipboard items by default.

Users could also use a hint about how to cleanly override environment
variables with systemd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,34 @@ there, but it basically works like this:
 2. If `clipmenud` detects changes to the clipboard contents, it writes them out
    to the cache directory.
 
+### Features of `clipmenud`
+
+The behavior of `clipmenud` can be customized through environment variables. Features include:
+
+ * Customizing max number of clips (Default: 1000)
+ * Choosing which selections to manage
+ * Ignoring certain windows, like password managers
+ * Enabling debugging
+ * Customizing the cache dir location
+ * Disable looping
+ * Option to "own" the clipboard
+
+Check the online help to view the details:
+
+    clipmenud --help
+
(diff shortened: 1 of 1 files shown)


cdown:
Hmm, I've added a note recently about `clipmenud --help`, but I'd rather not duplicate systemd documentation here, and I'd rather keep the readme shorter if possible. Could you go over more about the rationale to add this here as opposed to just pointing to `--help`?

markstos:
Each of these options is essentially a feature of `clipmenu`.

The ability to limit your history is particularly important feature for privacy.

A normal thing prospective users want to do when they are shopping for potential software to use is to be able to read a summary of the features offered to know whether they should even bother trying to install the software.

Put another way: Why pointing to the online help is OK for existing users to find the documentation, it's valuable for potential users to find this documentation without first installing the software.

Personally, a "max clips" feature is a key criteria on my shopping list for a clipboard manager. I imagine the other options exist because they are important to someone else who was shopping for a clipboard manager.

cdown:
> A normal thing prospective users want to do when they are shopping for potential software to use is to be able to read a summary of the features offered to know whether they should even bother trying to install the software.

Sure, then feel free to list some of the important features in a terse way, but we should just point to `clipmenud --help` to show the actual options -- let's avoid splitting sources of truth on documentation :-)

markstos:
I've now simplified the options docs and re-framed them as a terse description of "features".

cdown:
lgtm, thanks :-)

PR #124, opened 2020-04-08
Added entry to README.md for automatic pasting with sxhkd

jpnadas:
Simply setting CM_OUTPUT_CLIP=1 doesn't work with `sxhkd`. I managed to get it working with `xdotool`, so I added it to `README.md`, as I imagine this is something many people will want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,16 @@ package called `clipmenu`.
 If your distribution doesn't provide a package, you can run the scripts
 standalone (or better yet, package them!).
 
+## Configuration
+### Automatic pasting with sxhkd
+To get a `sxhkd` key binding to paste automatically, you can use `xdotool` to simulate the pressing of the paste shortcut.
+
+```
+# sxhkdrc
+alt + v
+    clipmenu && xdotool key Shift+Insert
+```
+
 # How does it work?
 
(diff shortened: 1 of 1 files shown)


cdown:
Thanks, but this seems highly specific (eg. most clipmenu are not also sxhkd users...), so I'm not in favour of adding this to the readme.

CM_OUTPUT_CLIP is about putting the clipboard contents on stdout anyway, I'm not sure how it's related.

jpnadas:
I think I confused putting it to stdout to pasting it... that is my bad.

In any case, don't you think this is relevant? sxhkd is widely used, and I think the behavior of pasting it automatically is quite nice to have.

cdown:
Where's the evidence that sxhkd is widely used by clipmenu users? There's only a single other issue or pull request that refers to it in any fashion (https://github.com/cdown/clipmenu/issues?q=sxhkd). If we added instructions for every piece of software which is "widely used", the readme would become an unending mess.

A generic guide to programmatically triggering a paste seems an independent piece of information from clipmenu operation, though, even if it's not about sxhkd.

jpnadas:
Ok, you convinced me. My intention was to save some time for people who tried to do the same thing as I did, but you are right, the readme is not the right place to have it.

I will withdraw the PR

PR #134, opened 2020-07-23
Run clipmenud after the display manager is started

nenad:
In cases of starting the display manager manually from tty, the DISPLAY variable is not yet set and clipmenud is crashing.


cdown:
Hi! This doesn't appear to be a standard unit, so you're probably going to have to add it as a drop-in.

Another way to provide `$DISPLAY` if X is not started yet is with `systemctl --user import-environment`.

nenad:
You are right. Turns out this didn't even fix it on my machine as I'm starting `sway` by executing `sway` from the tty and not through systemd. I'll close the PR, thanks for the input!

PR #137, opened 2020-08-16
Use xdotool instead of cat when CM_OUTPUT_CLIP

mil-ad:
This allows pasting directly in other non-terminal programs. For example a browser's address bar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # `dmenu` is not a hard dependency, but you need it unless
 # you plan to set CM_LAUNCHER to another value like `rofi`
-REQUIRED_BINS := xsel clipnotify
+REQUIRED_BINS := xsel xdotool clipnotify
 PREFIX ?= /usr
 $(foreach bin,$(REQUIRED_BINS),\
     $(if $(shell command -v $(bin) 2> /dev/null),$(info Found `$(bin)`),$(error Missing Dep. Please install `$(bin)`)))
--- a/clipmenu
+++ b/clipmenu
@@ -62,5 +62,5 @@ for selection in clipboard primary; do
 done
 
 if (( CM_OUTPUT_CLIP )); then
-    cat "$file"
+    xdotool key Shift+Insert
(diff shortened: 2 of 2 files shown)


cdown:
Hi there! Please maintain backwards compatibility of options :-)

It's also not acceptable to make xdotool mandatory, sorry. Please see how we already handle similar cases elsewhere.

cdown:
Closing without prejudice due to lack of response to feedback.

tblancher:
I wouldn't do this in clipmenu itself.  You should be able to assign this to however you launch clipmenu in your WM or DE.  For XMonad, I linked the following to my desired keybinding:

`spawn "clipmenu && xdotool key Ctrl+v"`

Replace "Ctrl+v" with whichever key binding you use for pasting.

johnwuelk:
Thanks! `... && xdotool key Ctrl+v` works flawless! I use xdotool anyway for various scripts.

nick87720z:
For case of interesting - text typing works with `xdotool type ...`, just needs '\r' replaced with '\r'. Though it may cause strange problem after use - when I did it against leafpad editor, it did not react to more key presses for short time, also Ctrl+Backspace did not work few times, when I used it to delete pasted text.

nick87720z:
I'm experimenting with clipmenu, trying to extend its own rofi-script mode with some operations. Of course, not sure if they are necessary to be right in that file.

This is snippet from `case $ROFI_RETV` block (in this case assigned to 2nd custom command):
```
11) chosen_line="$1"
        sed -E 's|(["\$])|\\\1|g
                s|$|\r|
                1s/^/xdotool type --clearmodifiers --delay 20 -- "/
                $s/$/"/' \
        < "${cache_dir}/$(printf '%s\n' "$chosen_line" | cksum)" | sh >/dev/null &
        exit;
```
Sure, 3rd mouse button simulation would be also useful. Or even more reliable (if any text is placed to selection - mouse paste works in far more cases).

PR #150, opened 2020-12-17
Preserve exit code from launcher

kankaristo:
This change is related to https://github.com/cdown/clipmenu/issues/57#issuecomment-740123283.

If `clipmenu` "preserves" the exit code from the launcher (exits with the same code as the launcher), the exit codes from Rofi's custom keybindings can be used with `clipmenu`.

For example, I'm using the following script bound to `Super+v`:
```bash
#!/bin/bash

trap "exit" INT

# Run clipmenu
CM_LAUNCHER=rofi clipmenu -p "Paste" -mesg "Use Shift+Delete to delete an item" \
    -kb-delete-entry "" -kb-custom-1 "Shift+Delete" \
    -kb-accept-alt "" -kb-custom-2 "Shift+Return"

exit_code=$?

case $exit_code in
    0) xdotool key "shift+Insert" ;;
    10) clipdel -d ^"$(xsel -b)"$; "$0" ;;
    *) exit $exit_code ;;
esac
```

With the above script, I have 3 different options when running `clipmenu`:
- if I press `Return`, the selected item is pasted to where my cursor currently is (a bit hackily with `xdotool`)
- if I press `Shift+Return`, the selected item is sent to the clipboard ("default" `clipmenu` behavior)
- if I press `Shift+Delete`, the selected item is deleted from `clipmenu`, and the script runs itself again (essentially keeps `clipmenu` open)

In order for this to work, the only change needed in `clipmenu` is to preserve the exit codes from the custom keybindings.

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -58,6 +58,8 @@ else
     chosen_line=$(list_clips | "$CM_LAUNCHER" "${launcher_args[@]}" "$@")
 fi
 
+launcher_exit=$?
+
 [[ $chosen_line ]] || exit 1
 file=$cache_dir/$(cksum <<< "$chosen_line")
 [[ -f "$file" ]] || exit 2
@@ -69,3 +71,5 @@ done
 if (( CM_OUTPUT_CLIP )); then
     cat "$file"
 fi
+
+exit $launcher_exit
--- a/clipmenu
+++ b/clipmenu
@@ -56,9 +56,8 @@ if [[ "$CM_LAUNCHER" == rofi-script ]]; then
     fi
 else
     chosen_line=$(list_clips | "$CM_LAUNCHER" "${launcher_args[@]}" "$@")
+    launcher_exit=$?
 fi
-
-launcher_exit=$?
 
 [[ $chosen_line ]] || exit 1
 file=$cache_dir/$(cksum <<< "$chosen_line")
@@ -72,4 +71,4 @@ if (( CM_OUTPUT_CLIP )); then
     cat "$file"
 fi
 
-exit $launcher_exit
+[ -n "$launcher_exit" ] && exit "$launcher_exit"
--- a/clipmenu
+++ b/clipmenu
@@ -71,4 +71,4 @@ if (( CM_OUTPUT_CLIP )); then
     cat "$file"
 fi
 
-[ -n "$launcher_exit" ] && exit "$launcher_exit"
+exit "${launcher_exit:-"$?"}"


clipmenu
```diff
@@ -58,6 +58,8 @@ else
     chosen_line=$(list_clips | "$CM_LAUNCHER" "${launcher_args[@]}" "$@")
 fi
 
+launcher_exit=$?
```
cdown, line 61: That's not quite right in the case at line 52. Shouldn't this be in the `else` block?

clipmenu
```diff
@@ -69,3 +71,5 @@ done
 if (( CM_OUTPUT_CLIP )); then
     cat "$file"
 fi
+
+exit $launcher_exit
```
cdown, line 75: It doesn't matter, but to avoid the wrath of the linter, quote this please :-)

cdown (changes requested):

cdown:
Looks good, thanks! Just a couple of nits.

clipmenu
kankaristo: If it's in the else block, and the variable on the last line is quoted, the last line would become `exit ""`, which causes an error `exit: numeric argument required`.

So, either this can't be in the else block, or the variable on the last line shouldn't be quoted (if it's not quoted, the line becomes `exit `, which is fine). I'm fine with doing it either way. :slightly_smiling_face: 

Yet another option would be to put the variable in the else block, and at the end check if the variable is empty, and only exit with it in that case?

clipmenu
kankaristo: ShellCheck was fine with it like this, but I'll quote just in case. :smile:

But if the variable is empty, quoting will produce an error in this case (see my comment on the other change).

kankaristo:
@cdown, the requested changes as they are could lead to an error (`exit ""`, which gives an error), so I've asked for more comments.

clipmenu
kankaristo: Moved the variable under the else statement.

clipmenu
kankaristo: I added quoting around the variable, and a "not-empty" check before calling exit, so the quoting won't cause an errror.

kankaristo:
@cdown, I've implemented the suggested changes, and added a check so that the launcher exit code is used only if the variable is set.

clipmenu
```diff
@@ -69,3 +70,5 @@ done
 if (( CM_OUTPUT_CLIP )); then
     cat "$file"
 fi
+
+[ -n "$launcher_exit" ] && exit "$launcher_exit"
```
cdown, line 74: Just write `exit "${launcher_exit:-"$?"}"` :-)

kankaristo: Sure, that works too, changed it. :smile:

cdown:
Looks good, thanks.

Issue #151, opened 2021-01-19
Status and toggle command


AckslD:
I've on my own setup extended `clipctl` to also have the commands:
* `status`: Whether disabled or enabled.
* `toggle`: Toggle between disabled and enabled.

It's currently a bit hacky but if you find this interesting and I can open a PR to add these commands and discuss how to do this in the best way.

cdown:
I think `status` sounds fine, but what's the use case for `toggle`? :-)

AckslD:
In my case I put an icon in my status bar showing it's status and I can either click on it or using a key-combination to toggle the status. This might be a specialised need though and a toggle command can easily be created if `status` exists.

cdown:
Sounds reasonable, in which case I think both are fine. Feel free to send a PR :-)

AckslD:
Cool, I might cleanup some things first and open a PR in the coming days :+1:

AckslD:
@cdown See PR #152. Let me know what you think :)

PR #152, opened 2021-01-20
Added commands to clipctl to check the status, toggle, get version and directory

AckslD:
This adds four new commands to `clipctl`, solving #151:
* `status`: returns "enabled" or "disabled"
* `toggle`: toggles clip collection
* `version`: returns major version
* `cm-dir`: returns the directory used for caching

I also changed other files to get the cache directory from `clipctl cm-dir`. I found this nicer to have it in one place, rather than having to have the correct folder set in each file. However, this is not required and I can revert that if you don't like it.

### DIFF
--- a/clipctl
+++ b/clipctl
@@ -1,27 +1,64 @@
 #!/usr/bin/env bash
 
+: "${CM_DIR:="${XDG_RUNTIME_DIR-"${TMPDIR-/tmp}"}"}"
+
 if [[ -z $1 ]] || [[ $1 == --help ]] || [[ $1 == -h ]]; then
     cat << 'EOF'
 clipctl provides controls for the clipmenud daemon.
 
 You can temporarily disable clip collection without stopping clipmenud entirely
 by running "clipctl disable". You can then reenable with "clipctl enable".
+
+Commands:
+    enable: enable clip collection
+    disable: disable clip collection
+    status: returns "enabled" or "disabled"
+    toggle: toggles clip collection
+    version: returns major version
+    cm-dir: returns the directory used for caching
 EOF
     exit 0
 fi
 
+_MAJORVERSION=6
 _CLIPMENUD_PID=$(pgrep -u "$(id -u)" -nf 'clipmenud$')
+_CLIPMENUCACHEDIR=$CM_DIR/clipmenu.$_MAJORVERSION.$USER
+_CLIPMENUSTATUSFILE=$_CLIPMENUCACHEDIR/status
 
-if [[ -z "$_CLIPMENUD_PID" ]]; then
-    echo "clipmenud is not running"
-    exit 2
-fi
+case $1 in
+    enable|disable|toggle)
+        if [[ -z "$_CLIPMENUD_PID" ]]; then
+            echo "clipmenud is not running"
+            exit 2
+        fi
+        ;;
+esac
 
 case $1 in
     enable) kill -USR2 "$_CLIPMENUD_PID" ;;
     disable) kill -USR1 "$_CLIPMENUD_PID" ;;
+    status)
+        if ! [[ $(pgrep -f "clipmenud") ]]; then
+            echo "disabled"
+            exit
+        fi
+        if [[ -e $_CLIPMENUSTATUSFILE ]]; then
+            echo $(cat $_CLIPMENUSTATUSFILE)
+        else
+            echo "enabled"
+        fi
+        ;;
+    toggle)
+        if [[ $(clipctl status) == "enabled" ]]; then
+            clipctl disable
+        else
+            clipctl enable
+        fi
+        ;;
+    version) echo $_MAJORVERSION ;;
+    cm-dir) echo $_CLIPMENUCACHEDIR ;;
     *)
         printf 'Unknown command: %s\n' "$1"
         exit 1
-    ;;
+        ;;
 esac
--- a/clipctl
+++ b/clipctl
@@ -15,50 +15,41 @@ Commands:
     status: returns "enabled" or "disabled"
     toggle: toggles clip collection
     version: returns major version
-    cm-dir: returns the directory used for caching
+    cache-dir: returns the directory used for caching
 EOF
     exit 0
 fi
 
-_MAJORVERSION=6
-_CLIPMENUD_PID=$(pgrep -u "$(id -u)" -nf 'clipmenud$')
-_CLIPMENUCACHEDIR=$CM_DIR/clipmenu.$_MAJORVERSION.$USER
-_CLIPMENUSTATUSFILE=$_CLIPMENUCACHEDIR/status
+clipmenud_pid=$(pgrep -u "$(id -u)" -nf 'clipmenud$')
 
 case $1 in
-    enable|disable|toggle)
-        if [[ -z "$_CLIPMENUD_PID" ]]; then
+    enable|disable|toggle|status)
+        if [[ -z "$clipmenud_pid" ]]; then
             echo "clipmenud is not running"
             exit 2
         fi
         ;;
 esac
 
+major_version=6
+cache_dir=$CM_DIR/clipmenu.$major_version.$USER
+status_file=$cache_dir/status
+
 case $1 in
-    enable) kill -USR2 "$_CLIPMENUD_PID" ;;
-    disable) kill -USR1 "$_CLIPMENUD_PID" ;;
-    status)
-        if ! [[ $(pgrep -f "clipmenud") ]]; then
-            echo "disabled"
-            exit
-        fi
-        if [[ -e $_CLIPMENUSTATUSFILE ]]; then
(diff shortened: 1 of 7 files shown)


clipctl
```diff
@@ -1,27 +1,64 @@
...
+            exit 2
+        fi
+        ;;
+esac
 
 case $1 in
     enable) kill -USR2 "$_CLIPMENUD_PID" ;;
     disable) kill -USR1 "$_CLIPMENUD_PID" ;;
+    status)
+        if ! [[ $(pgrep -f "clipmenud") ]]; then
```
cdown, line 41: If this is kept, I guess this can just do `[[ $_CLIPMENUD_PID ]]`?

clipctl
```diff
@@ -1,27 +1,64 @@
...
 You can temporarily disable clip collection without stopping clipmenud entirely
 by running "clipctl disable". You can then reenable with "clipctl enable".
+
+Commands:
+    enable: enable clip collection
+    disable: disable clip collection
+    status: returns "enabled" or "disabled"
+    toggle: toggles clip collection
+    version: returns major version
+    cm-dir: returns the directory used for caching
```
cdown, line 18: Oh, I like this and don't know why it never occurred to me that this might be a good place to centralise. Thanks!

clipctl
```diff
@@ -1,27 +1,64 @@
...
 EOF
     exit 0
 fi
 
+_MAJORVERSION=6
```
cdown, line 23: I don't remember why I wrote all these prefixed with an understore and in caps, but I guess I must have been high as a kite. Feel free to either change it to match the format from other files now, or I can just change it later :-)

clipctl
```diff
@@ -1,27 +1,64 @@
...
 _CLIPMENUD_PID=$(pgrep -u "$(id -u)" -nf 'clipmenud$')
+_CLIPMENUCACHEDIR=$CM_DIR/clipmenu.$_MAJORVERSION.$USER
+_CLIPMENUSTATUSFILE=$_CLIPMENUCACHEDIR/status
 
-if [[ -z "$_CLIPMENUD_PID" ]]; then
-    echo "clipmenud is not running"
-    exit 2
-fi
+case $1 in
+    enable|disable|toggle)
```
cdown, line 29: I think probably include `status` here, since "not running" is distinct from disabled.

clipctl
```diff
@@ -1,27 +1,64 @@
...
 
 case $1 in
     enable) kill -USR2 "$_CLIPMENUD_PID" ;;
     disable) kill -USR1 "$_CLIPMENUD_PID" ;;
+    status)
+        if ! [[ $(pgrep -f "clipmenud") ]]; then
+            echo "disabled"
+            exit
+        fi
+        if [[ -e $_CLIPMENUSTATUSFILE ]]; then
```
cdown, line 45: Needs quotes (RHS on `-e` undergoes expansion).

clipctl
```diff
@@ -1,27 +1,64 @@
...
+            echo $(cat $_CLIPMENUSTATUSFILE)
```
cdown, line 46: Wait, why both `echo` and `cat`? `cat` is enough :-)

clipctl
```diff
@@ -1,27 +1,64 @@
...
+            clipctl enable
+        fi
+        ;;
+    version) echo $_MAJORVERSION ;;
+    cm-dir) echo $_CLIPMENUCACHEDIR ;;
     *)
         printf 'Unknown command: %s\n' "$1"
         exit 1
-    ;;
+        ;;
```
cdown, line 63: Just preference, but I'd generally prefer these line up with the outer scope.

clipmenud
```diff
@@ -13,9 +12,10 @@ CM_MAX_CLIPS_THRESH=$(( CM_MAX_CLIPS + 10 ))
 : "${CM_SELECTIONS:=clipboard primary}"
 read -r -a selections <<< "$CM_SELECTIONS"
 
-major_version=6
-cache_dir=$CM_DIR/clipmenu.$major_version.$USER/
+cache_dir=$(clipctl cm-dir)
 cache_file=$cache_dir/line_cache
+status_file=$cache_dir/status
+echo "enabled" > $status_file
```
cdown, line 18: RHS needs quoting (in general, you'll benefit from running `shellcheck` before submission :-))

clipmenud
```diff
@@ -60,6 +60,7 @@ sig_disable() {
     info "Received disable signal, suspending clipboard capture"
     _CM_DISABLED=1
     _CM_FIRST_DISABLE=1
+    echo "disabled" > $status_file
```
cdown, line 63: ditto

clipmenud
```diff
@@ -78,6 +79,7 @@ sig_enable() {
 
     info "Received enable signal, resuming clipboard capture"
     _CM_DISABLED=0
+    echo "enabled" > $status_file
```
cdown, line 82: ditto

cdown (changes requested):
Thanks! Concept looks great, just some comments.

clipctl
cdown: Only mild comment, I think I prefer "cache-dir" since it's what we call it everywhere else.

clipctl
AckslD: Removed this, will be caught before now.

clipctl
AckslD: Changed to `cache-dir`

clipctl
AckslD: Hehe :) Changed to lowercase and without underscore

clipctl
AckslD: Added `status`

clipctl
AckslD: Removed this `if` statement completely. Since we now assert before that `clipmenud` is actually running there will for sure be such a file

clipctl
AckslD: Good point! :) Fixed

clipctl
AckslD: Fixed

clipmenud
AckslD: Added quoting. Why is quoting needed here actually?

Also, is there something misspelled? :)

clipmenud
AckslD: Done

clipmenud
AckslD: Done

clipctl
```diff
@@ -1,25 +1,53 @@
 #!/usr/bin/env bash
 
+: "${CM_DIR:="${XDG_RUNTIME_DIR-"${TMPDIR-/tmp}"}"}"
+
 if [[ -z $1 ]] || [[ $1 == --help ]] || [[ $1 == -h ]]; then
     cat << 'EOF'
 clipctl provides controls for the clipmenud daemon.
 
 You can temporarily disable clip collection without stopping clipmenud entirely
```
cdown, line 9: nit: Maybe remove this block since you now enumerate them all anyway.

clipctl
```diff
@@ -1,25 +1,53 @@
...
+major_version=6
+cache_dir=$CM_DIR/clipmenu.$major_version.$USER
+status_file=$cache_dir/status
 
 case $1 in
-    enable) kill -USR2 "$_CLIPMENUD_PID" ;;
-    disable) kill -USR1 "$_CLIPMENUD_PID" ;;
+    enable) kill -USR2 "$clipmenud_pid" ;;
+    disable) kill -USR1 "$clipmenud_pid" ;;
+    status) cat $status_file ;;
```
cdown, line 41: This one should also be quoted, since otherwise things will go wonky if `$CM_DIR` has spaces or similar.

clipctl
```diff
@@ -1,25 +1,53 @@
...
+    toggle)
+        if [[ $(clipctl status) == "enabled" ]]; then
+            clipctl disable
+        else
+            clipctl enable
+        fi
+    ;;
+    version) echo $major_version ;;
```
cdown, line 49: Just to avoid the shellcheck lottery which varies between versions: quote this :-)

clipctl
```diff
@@ -1,25 +1,53 @@
...
+    cache-dir) echo $cache_dir ;;
```
cdown, line 50: This also legitimately might fail for weird `$CM_DIR`s, quote it

cdown (changes requested):
Almost there. Thanks for working on this!

clipctl
AckslD: Done

clipctl
AckslD: Done

clipctl
AckslD: Done

clipctl
AckslD: Done

cdown (approved):

cdown:
Looks great, thank you!

AckslD:
My pleasure :)

PR #153, opened 2021-01-28
Allow multi select

deanbarrow:
This PR allows you to `Shift+Enter` to select multiple items to insert into the clipboard.

Useful for copying multiple items from a page and pasting them together in one go.

Credit to https://unix.stackexchange.com/a/140729 for `awk 'NR>1{print PREV} {PREV=$0} END{printf("%s",$0)}'`

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -38,7 +38,7 @@ if [[ "$CM_LAUNCHER" == fzf ]]; then
 fi
 
 # rofi supports dmenu-like arguments through the -dmenu flag
-[[ "$CM_LAUNCHER" == rofi ]] && set -- -dmenu "$@"
+[[ "$CM_LAUNCHER" == rofi ]] && set -- -dmenu -multi-select "$@"
 
 list_clips() {
     LC_ALL=C sort -rnk 1 < "$cache_file" | cut -d' ' -f2- | awk '!seen[$0]++'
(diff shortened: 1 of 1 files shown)


clipmenu
```diff
@@ -57,11 +57,25 @@ else
...
+
+files=()
+while IFS= read -r line; do
+    file=$cache_dir/$(cksum <<< "$line")
+    files+=("$file")
+done <<< "$chosen_line"
+
+paste=()
+for file in "${files[@]}"; do
+    contents=$(cat "$file")
```
cdown, line 69: Command substitution mangles cases with trailing newlines -- take a look at how this was covered in clipmenud for xsel.

clipmenu
```diff
@@ -57,11 +57,25 @@ else
...
+    paste+=("$contents")
```
cdown, line 70: Why do all this, why not actually use `cat` to concat and send it directly to xsel?

clipmenu
```diff
@@ -57,11 +57,25 @@ else
...
+done
+
+paste=$( IFS=$'\n'; echo "${paste[*]}" )
+
+ [[ $paste ]] || exit 2
 
 for selection in clipboard primary; do
-    xsel --logfile /dev/null -i --"$selection" < "$file"
+    echo "$paste" | awk 'NR>1{print PREV} {PREV=$0} END{printf("%s",$0)}' | xsel --logfile /dev/null -i --"$selection"
```
cdown, line 78: Wait, why are you explicitly mangling trailing newlines? That's literally something we try as hard as possible to preserve, so doing that is a non-starter.

clipmenu
```diff
@@ -57,11 +57,25 @@ else
...
+    files+=("$file")
+done <<< "$chosen_line"
+
+paste=()
+for file in "${files[@]}"; do
+    contents=$(cat "$file")
+    paste+=("$contents")
+done
+
+paste=$( IFS=$'\n'; echo "${paste[*]}" )
```
cdown, line 73: This just replaces `${paste[0]}` and leaves everything else intact. Either way, it seems unnecessary -- just use cat to join the data.

cdown (changes requested):
Hi! I'm afraid I'm a bit confused by what you're trying to achieve here. Could you please clarify your intentions?

clipmenu
deanbarrow: I tried with cat but it didn't work, I'll have another go as it would simplify the code a lot.

clipmenu
deanbarrow: This was specifically to remove the trailing newline that `echo` creates. I'm not sure if `cat` would behave the same but I'll test it.

clipmenu
deanbarrow: Will take a look.

clipmenu
deanbarrow: Will check, thanks.

cdown:
Closing without prejudice due to lack of update, feel free to resubmit. Thanks.

PR #159, opened 2021-03-14
Add backup functionality

korikori:
I needed a way to make (part of) my clipboard permanent, so I added a rudimentary backup function to clipctl and a handler for the CM_BACKUP environmental variable (full path to backup directory) in clipmenud.

### DIFF
--- a/clipctl
+++ b/clipctl
@@ -13,6 +13,7 @@ Commands:
     toggle: toggles clip collection
     version: returns major version
     cache-dir: returns the directory used for caching
+    backup: copy the current clip collection to a user-defined location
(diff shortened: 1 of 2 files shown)


cdown:
Hi! Sorry, but I think I'm unlikely to merge this -- it seems pretty niche.

clipmenu-agnostically, it should be possible to do something like the following:

```
for file in some_dir/*; do
    xclip -sel clip < "$file"

    # clipboard notifications in X are async, so wait a bit to avoid drops
    sleep 1
done
```

korikori:
Thank you for the input!

PR #161, opened 2021-03-25
Add ability to bypass clipmenu for choosing selection

simonhughxyz:
Add 2 new options:
1) --print-sel argument to print all selections in clipmenu
2) CM_INPUT_CLIP to accept selection from stdin

The reasoning behind this is to bypass clipmenu for getting a selection.
For example: `clipmenu --print-sel | fzf | CM_INPUT_CLIP=1 clipmenu`

For my user case I found a need for it so I can preprocess the output of fzf before piping it back into clipmenu.
This is because fzf prints out extra information, apart from the selection, in some circumstances and clipmenu can't handle with such a scenario.

### DIFF
--- a/clipmenu
+++ b/clipmenu
@@ -50,7 +50,8 @@ fi
 launcher_args=(-l "${CM_HISTLENGTH}")
 
 if (( CM_INPUT_CLIP )); then
-    chosen_line="$( cat )"
+    read
+    chosen_line="$REPLY"
 else
     # Blacklist of non-dmenu launchers
     launcher_args=(-l "${CM_HISTLENGTH}")
--- a/clipmenu
+++ b/clipmenu
@@ -47,23 +47,30 @@ fi
 
 # Blacklist of non-dmenu launchers
 launcher_args=(-l "${CM_HISTLENGTH}")
-if [[ "$CM_LAUNCHER" == fzf ]]; then
-    launcher_args=()
-fi
 
-# rofi supports dmenu-like arguments through the -dmenu flag
-[[ "$CM_LAUNCHER" == rofi ]] && set -- -dmenu "$@"
-
-if [[ "$CM_LAUNCHER" == rofi-script ]]; then
-    if (( $# )); then
-        chosen_line="${!#}"
-    else
-        list_clips
-        exit
-    fi
+if (( CM_INPUT_CLIP )); then
+    chosen_line="$( cat )"
 else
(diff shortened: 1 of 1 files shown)


clipmenu
```diff
@@ -22,9 +21,25 @@ Environment variables:
...
+
+print_sel(){
+    list_clips
+    exit 0
+}
+
+case "$1" in
+    # Not -h, see #142
+    --help) print_help ;;
+    --print-sel) print_sel ;;
```
cdown, line 41: Hmm, why not implement `CM_LAUNCHER=cat` instead in a similar manner to fzf?

clipmenu
```diff
@@ -33,27 +48,30 @@ fi
...
 # Blacklist of non-dmenu launchers
 launcher_args=(-l "${CM_HISTLENGTH}")
-if [[ "$CM_LAUNCHER" == fzf ]]; then
-    launcher_args=()
-fi
 
-# rofi supports dmenu-like arguments through the -dmenu flag
-[[ "$CM_LAUNCHER" == rofi ]] && set -- -dmenu "$@"
+if (( CM_INPUT_CLIP )); then
+    chosen_line="$( cat )"
```
cdown, line 53: Just `read` -- it can't be more than one line anyway, so no need to fork `cat`. :-)

cdown (changes requested):
Seems reasonable, but `CM_LAUNCHER=cat` with a similar case to `fzf` would be ideal. Thanks!

clipmenu
simonhughcom: Could do, But upon reviewing this code a bit more, I think I have a much better way of resolving this problem but it will need a brand new PR.
I'll quickly fix this, and when I have time I will post the new PR and see what you think.

clipmenu
```diff
@@ -33,27 +42,34 @@ fi
...
+    [[ "$CM_LAUNCHER" == rofi ]] && set -- -dmenu "$@"
+
+    if [[ "$CM_LAUNCHER" == rofi-script ]]; then
+        if (( $# )); then
+            chosen_line="${!#}"
+        else
+            list_clips
+            exit
+        fi
+    elif [[ "$CM_LAUNCHER" == cat ]]; then
```
cdown, line 66: Hmm, why not just use the same logic as fzf and actually use `CM_LAUNCHER`?

simonhughcom: You lost me, that's what I did?

Also, could you review the other PR I did, which has, under my view, a much better solution to this problem.

cdown:

PR #162, opened 2021-03-26
Add new clipcli to seperate clipboard management from dmenu

simonhughxyz:
So this is my improved solution to a a problem I solved in #161 

I have added a new script called `clipcli` which handles the clipboard management.
It has 2 commands
- `clipcli list`: prints a list of clips.
- `clipcli get`: retrieves a clip based on its input.

So you can build your own scripts using `clipcli list | fzf | clipcli get`

I think this is a better solution because I believe that the clipboard management should be independent of whatever menu of fuzzy finder you use.

I have converted clipmenu to use the new script but kept the cli of clipmenu the same for backwards compatibility, but I think it may be a good idea to depreciate it and introduce separate scripts for dmenu, fzf, rofi etc

This would allow people to build their own scripts more easily, as a possible suggestion you could have a new dir in this repo or another repo with user scrips people can publish.

This was made in a bit of a rush, so I might have missed some things.

### DIFF
--- a/clipcli
+++ b/clipcli
@@ -41,8 +41,8 @@ get_clip(){
 case "$1" in
     # Not -h, see #142
     -h|--help) print_help ;;
-    -l|--list) list_clips; exit 0 ;;
-    -g|--get) get_clip "$2"; exit 0 ;;
+    list) list_clips; exit 0 ;;
+    get) get_clip "$2"; exit 0 ;;
 esac
 
 
--- a/clipcli
+++ b/clipcli
@@ -32,7 +32,12 @@ list_clips() {
 }
 
 get_clip(){
-    [ $1 ] && clip_line="$1" || read; clip_line="$REPLY"
+    if [[ -n $1 ]]; then
+        clip_line="$1"
+    else
+        read;
+        clip_line="$REPLY"
+    fi
     file=$cache_dir/$(cksum <<< "$clip_line")
     [[ -f "$file" ]] || exit 2
     cat "$file"
--- a/clipcli
+++ b/clipcli
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# CLI for listing and retrieving clips
+
+: "${CM_DIR="${XDG_RUNTIME_DIR-"${TMPDIR-/tmp}"}"}"
+
+major_version=6
+
+shopt -s nullglob
+
+cache_dir=$CM_DIR/clipmenu.$major_version.$USER
+cache_file=$cache_dir/line_cache
+
+if ! [[ -f "$cache_file" ]]; then
+    printf '%s\n' 'No cache file yet, did you run clipmenud?'
+    exit 2
+fi
+
+print_help(){
+    cat << 'EOF'
+clipmenu is a simple clipboard manager using dmenu and xsel. Launch this
+when you want to select a clip.
+All arguments are passed through to dmenu itself.
+Environment variables:
+- $CM_DIR: specify the base directory to store the cache dir in (default: $XDG_RUNTIME_DIR, $TMPDIR, or /tmp)
+EOF
+    exit 0
+}
+
+list_clips() {
+    LC_ALL=C sort -rnk 1 < "$cache_file" | cut -d' ' -f2- | awk '!seen[$0]++'
+}
+
+get_clip(){
+    [ $1 ] && clip_line="$1" || read; clip_line="$REPLY"
+    file=$cache_dir/$(cksum <<< "$clip_line")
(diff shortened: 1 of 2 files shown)


clipcli
```diff
@@ -0,0 +1,53 @@
...
+cache_file=$cache_dir/line_cache
+
+if ! [[ -f "$cache_file" ]]; then
+    printf '%s\n' 'No cache file yet, did you run clipmenud?'
+    exit 2
+fi
+
+print_help(){
+    cat << 'EOF'
+clipmenu is a simple clipboard manager using dmenu and xsel. Launch this
```
cdown, line 21: This help looks wrong to me, since it references clipmenu, no?

clipcli
```diff
@@ -0,0 +1,53 @@
...
+EOF
+    exit 0
+}
+
+list_clips() {
+    LC_ALL=C sort -rnk 1 < "$cache_file" | cut -d' ' -f2- | awk '!seen[$0]++'
+}
+
+get_clip(){
+    if [[ -n $1 ]]; then
```
cdown, line 35: Just [[ $1 ]] is fine.

clipcli
```diff
@@ -0,0 +1,53 @@
...
+        clip_line="$1"
+    else
+        read;
```
cdown, line 38: Why is there a semicolon here? :-)

clipcli
```diff
@@ -0,0 +1,53 @@
...
+        clip_line="$REPLY"
```
cdown, line 39: You can just do `read clip_line`. You'll also want to use `IFS= read -r`.

clipcli
```diff
@@ -0,0 +1,53 @@
...
+    fi
+    file=$cache_dir/$(cksum <<< "$clip_line")
+    [[ -f "$file" ]] || exit 2
+    cat "$file"
+}
+
+case "$1" in
+    # Not -h, see #142
+    -h|--help) print_help ;;
+    list) list_clips; exit 0 ;;
```
cdown, line 49: Why explicit `exit 0`? Why not just preserve it?

clipmenu
```diff
@@ -40,32 +40,26 @@ fi
...
 else
-    chosen_line=$(list_clips | "$CM_LAUNCHER" "${launcher_args[@]}" "$@")
-    launcher_exit=$?
+    chosen_line=$(clipcli list | "$CM_LAUNCHER" "${launcher_args[@]}" "$@")
 fi
 
 [[ $chosen_line ]] || exit 1
-file=$cache_dir/$(cksum <<< "$chosen_line")
-[[ -f "$file" ]] || exit 2
+clip(){ clipcli get "$chosen_line"; }
```
cdown, line 55: `$chosen_line` comes implicitly, which makes this code hard to reason about for humans or static analysis. Just open code it.

clipmenu
```diff
@@ -40,32 +40,26 @@ fi
...
-    xsel --logfile /dev/null -i --"$selection" < "$file"
+    clip | xsel --logfile /dev/null -i --"$selection"
 done
 
 if (( CM_OUTPUT_CLIP )); then
-    cat "$file"
+    clip
 fi
 
 exit "${launcher_exit:-"$?"}"
```
cdown, line 65: Huh? You deleted `launcher_exit`, but it's still used

clipcli
```diff
@@ -0,0 +1,53 @@
...
+        read;
+        clip_line="$REPLY"
+    fi
+    file=$cache_dir/$(cksum <<< "$clip_line")
+    [[ -f "$file" ]] || exit 2
+    cat "$file"
+}
+
+case "$1" in
+    # Not -h, see #142
```
cdown, line 47: Huh? It's right below

cdown (changes requested):
Thanks for submitting, apologies for the delayed review. In general I think separating this is fine, but I'm a bit confused by some of the changes. Please also pass your changes through `shellcheck` and you'll see more of these issues ahead of time.

nick87720z:
Just curious, can't it be clipctl option, like `clipctl --stdin`, making it to perform commands from stdin? This would at least avoid duplication.

nick87720z:
My appologies - did not look commits before writing what I did before. Finally - what's really missing, but could be there.

I thought, clipcli takes commands from stdin coninuously, but it's not. Imho, persistant daemon would be useful, moreover it's shell script, whose runing creates significant overhead.

Also - the most effective way could be if actual job was done by `clipmenud`, providing named pipe to control itself (enable/disable commands could also use this interface). In this case there would be no possible race from different cli accessing cache - clipmenud would do everything.

cdown:

PR #171, opened 2021-11-19
clipmenud: fix killing background jobs by unquoting variable

mehw:
Hello,

I found out that a variable storing the multiline result of `jobs -p` should not be quoted when it's used as sequence of arguments.

Here I propose a trivial fix... to just unquote the variable.

This allows to properly kill all children jobs.

```
$ jobs -p
7345
7862

$ bg=$(jobs -p)

$ echo $bg
7345 7862

$ echo "$bg"
7345
7862

Correct:
$ kill -- $bg
$ kill -- 7345 7862

Wrong:
$ kill -- "$bg"
$ kill -- "7345"$'\n'"7862"
```

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -86,11 +86,11 @@ kill_background_jobs() {
     # group leader. As such, all we can do is look at the pending jobs. Bash
     # avoids a subshell here, so the job list is in the right shell.
     local bg
-    bg=$(jobs -p)
+    bg=( $(jobs -p) )
 
     # Don't log `kill' failures, since with KillMode=control-group, we're
     # racing with init.
-    [[ $bg ]] && kill -- $bg 2>/dev/null
+    [[ $bg ]] && kill -- ${bg[*]} 2>/dev/null
 }
 
 if [[ $1 == --help ]] || [[ $1 == -h ]]; then
--- a/clipmenud
+++ b/clipmenud
@@ -91,9 +91,7 @@ kill_background_jobs() {
 
     # Don't log `kill' failures, since with KillMode=control-group, we're
     # racing with init.
-    for i in ${!bg[@]}; do
-        kill -- ${bg[i]} 2>/dev/null
-    done
+    kill -- "${bg[@]}" 2>/dev/null
 }
 
 if [[ $1 == --help ]] || [[ $1 == -h ]]; then
--- a/clipmenud
+++ b/clipmenud
@@ -85,12 +85,15 @@ kill_background_jobs() {
     # While we usually _are_, there are no guarantees that we're the process
     # group leader. As such, all we can do is look at the pending jobs. Bash
     # avoids a subshell here, so the job list is in the right shell.
-    local bg
-    bg=( $(jobs -p) )
+    local -a bg
+
+    readarray -t bg < <(jobs -p)
 
     # Don't log `kill' failures, since with KillMode=control-group, we're
     # racing with init.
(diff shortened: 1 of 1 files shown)


cdown:
Thanks! This is indeed a bug, good catch.

In this case it makes no functional difference, but I'd prefer if this was done using an array instead of by field splitting, since it makes it clear what the intent is, so I guess `read -ra` to a local variable.

mehw:
> In this case it makes no functional difference, but I'd prefer if this was done using an array instead of by field splitting, since it makes it clear what the intent is, so I guess `read -ra` to a local variable.

Another way to use an array is by using round brackets and `*` as the expansion parameter.

```
$ bg=( $(jobs -p) )
$ echo ${bg[*]}
```

cdown:
Sure, but it again is done by implicit IFS expansion, and I'd rather avoid that at all as a stylistic choice.

mehw:
> Sure, but it again is done by implicit IFS expansion, and I'd rather avoid that at all as a stylistic choice. It doesn't matter here, but it makes it clearer to readers and static analysis that the splitting is intentional.

I experimented with `readarray` as an alternative solution. This implementation should be `IFS` agnostic.

Do you think it's better to explicitly use `IFS`?

PS: When you agree on a sulution I may squash the commits ;)

```
test_me() {
    local -a bg

    clipnotify &
    clipnotify &

    IFS=$'\t' # or anything else, just to try to mess things up...

    readarray -t bg < <(jobs -p)

    echo -n "\${bg[@]}: "
    echo ${bg[@]}

    echo -n "\${bg[*]}: "
    echo ${bg[*]}

    echo -n "\"\${bg[@]}\": "
    echo "${bg[@]}"

    echo -n "\"\${bg[*]}\": "
    echo "${bg[*]}"

    echo -n "\${!bg[@]}: "
    echo ${!bg[@]}

    echo -n "\${!bg[*]}: "
    echo ${!bg[*]}

    echo -n "\"\${!bg[@]}\": "
    echo "${!bg[@]}"

    echo -n "\"\${!bg[*]}\": "
    echo "${!bg[*]}"

    for i in ${!bg[@]}; do
	kill -- ${bg[i]}
    done
}

test_me
```

clipmenud
```diff
@@ -85,12 +85,15 @@ kill_background_jobs() {
...
+    local -a bg
+
+    readarray -t bg < <(jobs -p)
 
     # Don't log `kill' failures, since with KillMode=control-group, we're
     # racing with init.
-    [[ $bg ]] && kill -- "$bg" 2>/dev/null
+    for i in ${!bg[@]}; do
+        kill -- ${bg[i]} 2>/dev/null
+    done
```
cdown, line 96: Why do all this instead of just

    kill -- "${bg[@]}" 2>/dev/null

?

cdown:
Just the question about `"${bg[@]}"` to go, I can squash if you like on merge.

clipmenud
mehw: No particular reason... looking around in `clipmenud` I saw a `while` looping an array, so I kind of copycatted it...

mehw:
> Just the question about `"${bg[@]}"` to go, I can squash if you like on merge.

Possibly the last commit should fix it... Let me know.

Thanks.

cdown:
squashed/merged to bc35b6f44ab0def13601c4ebd836437c5dc0e360, thanks!

mehw:
Thank you for the patience fixing this PR ;)

PR #172, opened 2021-11-19
clipmenud: prevent spawning more clipnotify jobs

mehw:
Hello,

While `clipmenud` is running, each time `clipctl enable` is executed another `clipnotify` job is spawn...

This bug is how I found out the need of #171 (fix to properly kill children jobs).

```
$ clipmenud &
...
$ ps -e | grep clipnotify
7382 pts/1   00:00:00 clipnotify
$ clipctl enable
...
$ ps -e | grep clipnotify
7382 pts/1   00:00:00 clipnotify
7395 pts/1   00:00:00 clipnotify
```

There's a problem in the main loop, which is easily fixable... When a signal is trapped, the `wait "$_CM_CLIPNOTIFY_PID"` breaks, and the loop restarts, spawning another `clipnotify` job because the previous one was not killed by `sig_enable`...

On the other hand, `sig_disable` kills the current `clipnotify` job before restarting the loop... We need to do the same for `sig_enable`.

Thanks.

### DIFF
--- a/clipmenud
+++ b/clipmenud
@@ -58,9 +58,7 @@ debug() { (( CM_DEBUG )) && printf '%s\n' "$@" >&2; }
 sig_disable() {
     info "Received disable signal, suspending clipboard capture"
     _CM_DISABLED=1
-    _CM_FIRST_DISABLE=1
     echo "disabled" > "$status_file"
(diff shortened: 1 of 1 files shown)


cdown:
I'll look at this later, thanks.

cdown:
Thanks for reporting! This is indeed a bug, thanks for the fix. :-)

mehw:
Thanks for merging ;)

PR #173, opened 2021-11-19
clipctl: add start and stop clipmenud daemon options

mehw:
Hello,

This is to propose to add **start/stop clipmenud daemon** functionality to `clipctl`.

Let me know what you think.

Thanks.

### DIFF
--- a/clipctl
+++ b/clipctl
@@ -7,11 +7,11 @@ if [[ -z $1 ]] || [[ $1 == --help ]] || [[ $1 == -h ]]; then
 clipctl provides controls for the clipmenud daemon.
 
 Commands:
-    enable: enable clip collection
-    disable: disable clip collection
-    status: returns "enabled" or "disabled"
-    toggle: toggles clip collection
-    version: returns major version
+       enable: enable clip collection
+      disable: disable clip collection
+       status: returns "enabled" or "disabled"
+       toggle: toggles clip collection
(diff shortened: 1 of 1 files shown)


cdown:
Hmm, what's the use case? Ideally clipmenud should be managed by a service manager anyway, so unless there's some compelling reason I'd rather leave that to do it.

mehw:
This is to have manual control. Not everyone uses a service manager like `systemd`...

For instance, this is an universal way to start/stop the service. The daemon `clipmenud` is efficient, starting and stopping it should be practical too, without any overhead, and `clipctl` already presents itself as a way to control the daemon.

cdown:
If that's the only reason then I don't think it's particularly compelling, one might as well just run `pkill clipmenud` or start `clipmenud`. The only reason `-USR1` and `-USR2` have enable/disable is because the signal used is an implementation detail, but SIGTERM/SIGINT aren't.

cdown:
Closing this out without prejudice based on the above comment.

mehw:
> If that's the only reason then I don't think it's particularly compelling, one might as well just run `pkill clipmenud` or start `clipmenud`. The only reason `USR1` and `USR2` are mapped to disable/enable is because the signal used is an implementation detail, but SIGTERM/SIGINT aren't.

I agree. Thanks for the explanation.

PR #174, opened 2021-11-19
clipdel: allow ".*" to print the cache directory to be deleted

mehw:
Hello,

I find useful to print the cache directory that will be deleted by calling `clipdel ".*"` directly, rather than `clipctl cache-dir`, before doing an irreversible action via the `-d` option.

The user could use the pattern `"."` to print all the cached lines, and `".*"` to print the cache directory itself, before completing the operation with the `-d` option.

This PR also simplifies the code logic (nested conditions) of the `clipdel` script.

What do you think? Could this be a good idea?

### DIFF
--- a/clipdel
+++ b/clipdel
@@ -53,22 +53,24 @@ sed_common_command="s#^[0-9]\+ ##;\\#${esc_pattern}#"
 if ! [[ $raw_pattern ]]; then
     printf '%s\n' 'No pattern provided, see --help' >&2
     exit 2
-fi
-
-exec {lock_fd}> "$lock_file"
-
-if (( CM_REAL_DELETE )) && [[ "$raw_pattern" == ".*" ]]; then
-    flock -x -w "$lock_timeout" "$lock_fd" || exit
-    rm -rf -- "$cache_dir"
-    mkdir -p -- "$cache_dir"
-    exit 0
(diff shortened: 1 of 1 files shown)


cdown:
> I find useful to print the cache directory that will be deleted by calling clipdel ".*" directly, rather than clipctl cache-dir, before doing an irreversible action via the -d option.

Sorry, could you explain what you mean by that? Not totally sure I understand why one would prever one over the other :-)

mehw:
> Sorry, could you explain what you mean by that? Not totally sure I understand why one would prever one over the other :-)

IMHO it's more informing. If I'm thinking about `clipdel` to delete the cache directory, I would like to be informed which directory is deleted by `clipdel` itself, rather than switching to `clipctl`. This is also what `clipdel` does when a pattern to delete a line (aka clip) is given, so why not doing the same for the cache directory...

```
# Print the cache directory.
$ clipctl cache-dir
/tmp/1000-runtime-dir.aER/clipmenu.6.user

# Print what will be deleted.
$ clipdel ".*"
delete cache dir: /tmp/1000-runtime-dir.aER/clipmenu.6.user

# Confirm the deletion.
$ clipdel -d ".*"
```

cdown:
Sure, sounds reasonable then, thanks.

mehw:
Thanks ;)

PR #185, opened 2022-04-13
Die if X server is killed.

navneetankur:


navneetankur:
> Could you please explain your use case?

I don't always have my window manager running. If I close my window manager, clipmenud fills the terminal with error lines. Making the terminal unusable. Also, it uses high amount of CPU due to the loop continuously running.

> This adds a dependency on xset, which isn't acceptable just for this.

Okay, that makes sense.

Though I will add two things.
1. Maybe the xset check isn't needed at all. If clipnotify doesn't give a non zero exit code when things are fine, then only the exit code of clipnotify can be checked. What does it mean when clipnotify returns a non zero exit code?
2. If that's not enough, I think the output of clipnotify can be checked. It says something along the lines of 'can't connect to xserver'.
3. xset check doesn't have to add a dependency. The code will work fine without xset, if we add a check before the xset check.
    Something like, `if type xset`
    If xset is not available, this feature(code path) can be completely ignored.

navneetankur:
> At worst I think all we'd need is exponential backoff.

I guess that would help my use case too. Maybe it will allow enough time to use the tty to kill the clipmenud.

navneetankur:
> clipmenud is dependent on X, but shouldn't be dependent on having any window manager running.

Well, I close X too. Sorry for the confusion.

> I'm ok with adding a backoff. Exponential backoff seems excessive, but adding sleep 10 with a comment is fine.

Thanks.

clipmenud
```diff
@@ -152,6 +152,10 @@ while true; do
         clipnotify &
         _CM_CLIPNOTIFY_PID="$!"
         wait "$_CM_CLIPNOTIFY_PID"
+        if [[ $? -ne 0 ]]; then
```
cdown, line 155: No need to manually test $?, just do `wait ... || sleep 10` with the comment above. And it would be great to have a space between the octothorpe and the first word :-)

navneetankur: Done. Also, I learned a new word.

cdown:
Oh sorry, I realise I made an error in my last round of review. What we probably actually want is:

```
if ! wait "$_CM_CLIPNOTIFY_PID";
    # X server dead?
    sleep 10
    continue
fi
```

Sorry to ask for another change, that's my bad.

cdown:
Thank you! Looks great.

navneetankur:
Thanks

PR #186, opened 2022-05-05
Handle syncing clipboard to primary

bobrippling:
This mirrors behaviour in #126 / the reverse fix for #125, where if I copy in certain programs (google docs), the copy isn't synced to my primary selection. Usually this is implicit because copying requires highlighting, which sets the primary selection - possibly related the what was mentioned here - https://github.com/cdown/clipmenu/pull/126#issuecomment-719533069.

Understandably there are uses of CLIPBOARD distinct from PRIMARY, which is why I feel this should be behind a `CM_SYNC_CLIPBOARD_TO_PRIMARY` variable.


cdown:
Hi! Sorry to be late getting back. This looks fine, merging. Thanks!

PR #201, opened 2023-03-23
Improve test suite isolation, plus miscellaneous test suite changes

tomeon:
First, thanks for `clipmenu` -- it has upped my clipboard game significantly :)

While working on a different PR, I ran the `clipmenu` test suite and was surprised to find that it clobbered my clipboard cache.  This PR is my attempt to isolate the test suite from the user's existing state; the PR also introduces some miscellaneous additional changes.

Summary of changes:

1. Introduce a test fixtures hierarchy.  The fixtures directory contains a `bin` path with executables that replace the executables previously dynamically generated in `tests/test-clipmenu` and `tests/test-perf`.
2. Prepend the `test/fixtures/bin` directory and the `clipmenu` repository toplevel directory to `PATH`.  This makes it possible to run (say) `clipctl cache-dir` and be confident that the selected `clipctl` executable is `<toplevel>/clipctl`.
3. Put all generated assets (including the clipboard cache file) into a shared temporary directory (`tempdir=$(mktemp -d)`).  This simplifies end-of-test filesystem cleanup (just recursively remove `tempdir`).
4. Add a `cleanup` function that runs as a `trap` on the `EXIT` pseudo-signal.  This function recursively removes the above-mentioned `tempdir`, and also provides a `cleanup_pre` hook for running code prior to removing the temporary directory hierarchy.  `tests/test-clipmenu` uses `cleanup_pre` to print the contents of the temporary file used for capturing the mocked-up launchers' output.
5. Consolidated common test setup (including some of the things mentioned above) into the file `tests/setup.sh`, and source this file from `tests/test-clipmenu` and `tests/test-perf`.
6. In `tests/test-perf`, ensure that we always generate 1500 clipboard files.  Previously, if `NO_RECREATE` was in effect and a previous test run generated fewer that 1500 clipboard files (say, the test run was interrupted during the generation stage), subsequent runs with `NO_RECREATE` in effect would not generate the remaining needed files.
7. Print a  diagnostic message in the `cleanup` trap function when a test script failed.  The diagnostic message includes the command that failed, to aid in debugging.
8. Use a dynamically-assigned `BASH_XTRACEFD` in the `test/fixtures/bin/clipmenu-perf` wrapper rather than using the default `stderr` file descriptor.
9. Define a `PS4` that results in `xtrace` displaying additional execution context -- filename, line number, and (if applicable) shell function.

Thanks again for `clipmenu`, and thanks in advance for your consideration.

### DIFF
--- a/tests/fixtures/bin/clipmenu-launcher
+++ b/tests/fixtures/bin/clipmenu-launcher
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+# shellcheck disable=SC1091
+SHIM_STDOUT="Selected text. (2 lines)" source clipmenu-test-shim "$@"
--- a/tests/fixtures/bin/clipmenu-perf
+++ b/tests/fixtures/bin/clipmenu-perf
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
-exec 3>&2 2> >(tee "${log?}" |
+exec {BASH_XTRACEFD}> >(tee "${log?}" |
                  sed -u 's/^.*$/now/' |
                  date -f - +%s.%N > "${tim?}")
+
+export BASH_XTRACEFD
+
 set -x
 
 dmenu() { :; }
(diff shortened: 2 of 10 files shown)


cdown:

(note: rofi-script support is not in clipmenu 7, if that's a problem, let's talk about it)

PR #202, opened 2023-03-23
Improve launcher detection, and support a "custom" launcher type

tomeon:
Marked as a draft because it depends on changes from #201. 

Summary of changes:

1. Introduce support for a [custom launcher type](https://github.com/cdown/clipmenu/compare/develop...tomeon:extend-launcher-support?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R147-R168) that exposes an interface for the user to dynamically select a "real" launcher implementation like `dmenu` or `fzf` and execute the selected launcher with appropriate arguments.
2. Detect when `clipmenu` is being run as a rofi script by checking for the presence of the `ROFI_RETV` environment variable.
3. Document [the behavior that clipmenu expects from a launcher](https://github.com/cdown/clipmenu/compare/develop...tomeon:extend-launcher-support?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R71-R98).
4. Document [how to select a launcher](https://github.com/cdown/clipmenu/compare/develop...tomeon:extend-launcher-support?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R100-R145).

Note that this PR includes a few breaking changes, documented in the commits that introduced them:

- [741e785](https://github.com/cdown/clipmenu/commit/741e7857acbb58961805386a445e4e41858fa5ad)
- [16d207d](https://github.com/cdown/clipmenu/commit/16d207dc8d3cda0932a690b0cdd5b0065b5dda84)

Thanks for your consideration!

PR #203, opened 2023-03-23
rofi script mode options

tomeon:
Marked as a draft because it is based upon #202.

This PR updates `clipmenu` so that, when executed in `rofi-script` mode, it uses rofi's script option setting protocol to:

1. Update the `rofi` prompt to `clipmenu: `, and
2. Prevent custom input (that is, input that does not match an available selection) for the reasons explained in [the commit](https://github.com/cdown/clipmenu/commit/d59dcd0bda309e08cac69f86f4034befd8f7c2f6).

Thanks!


cdown:

(note: custom launchers are generally better supported with(out) CM_LAUNCHER_PASS_DMENU_ARGS now)

PR #211, opened 2024-01-05
Recognize any set CM_OUTPUT_CLIP value.

Labdabeta:
Instead of requiring CM_OUTPUT_CLIP to be specifically set to a non-zero integer, this allows CM_OUTPUT_CLIP to be set to any value, even an empty string.


cdown:

PR #219, opened 2024-06-06
allow selecting the default launcher at build time

N-R-K:
This will make things a bit easier to package on gentoo where we want to add the default launcher as a dependency. Currently, the ebuild simply runs sed over the bash script do accomplish it [^0].

[^0]: https://github.com/gentoo/gentoo/blob/1c154d09185a0b0911e9db506d6f00c7941759d2/x11-misc/clipmenu/clipmenu-6.2.0-r1.ebuild#L29-L33

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
+LAUNCHER := demnu
 CFLAGS := -std=gnu11 -O2 -Wall -Wextra -Wshadow -Wpointer-arith \
 	  -Wcast-align -Wmissing-prototypes -Wstrict-overflow -Wformat=2 \
 	  -Wwrite-strings -Warray-bounds -Wstrict-prototypes \
 	  -Wno-maybe-uninitialized \
-	  -Werror $(CFLAGS)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,8 +1,9 @@
+LAUNCHER := demnu
```
N-R-K, line 1: Went with the obvious name `LAUNCHER`, something more verbose like `DEFAULT_LAUNCHER` works for my use case too if that's preferred.

cdown:
I think configuration via the config file is fine, I'd rather not have a separate method for this one thing.

N-R-K:
> I think configuration via the config file is fine, I'd rather not have a separate method for this one thing.

That's reasonable. Selecting the default at build time is mostly so that the user can choose which launcher to pull in as a dependency. E.g if one prefers rofi then he can select it as default launcher and dmenu won't get pulled in anymore.

But that's very much gentoo specific thing. So we can just carry the patch there.

PR #220, opened 2024-06-06
readme: remove outdated information

N-R-K:
Ideally should be replaced with new information. But for now just removes the outdated stuff to avoid confusing potential new users since `develop` is set as the default branch.


cdown:
rofi-script is what I'm waiting on release for, I need to fix that up. Until then I'd rather not visibly show it as "removed", it's just pending.

N-R-K:
> rofi-script is what I'm waiting on release for

Is there a list of blocker for v7? So far what I've gathered are:

- [ ] rofi-script
- [ ] manpages
- [ ] readme update

Anything else?

PR #221, opened 2024-06-06
clipmenud: avoid spawning multiple daemons

N-R-K:
things can get messy if a running clipmenud is diabled via clipctl and then another clipmenud daemon is started. avoid spawning multiple daemons by using a session lock similar to the old bash script.

### DIFF
--- a/src/clipmenud.c
+++ b/src/clipmenud.c
@@ -10,6 +10,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/file.h>
 #include <sys/select.h>
 #include <sys/signalfd.h>
 #include <time.h>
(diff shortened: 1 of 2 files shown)


src/clipmenud.c
```diff
@@ -384,6 +385,14 @@ int main(void) {
     int evt_base;
 
     cfg = setup("clipmenud");
+
+    _drop_(close) int session_fd =
```
cdown, line 390: Probably also O_CLOEXEC?

cdown (commented):
Thanks!

src/clipmenud.c
N-R-K: Done. But does clipmenud do any `exec`? Also what about `content_dir_fd` and `snip_fd`, shouldn't they be `O_CLOEXEC`ed too if it's just out of precaution/principle?

cdown: `O_CLOEXEC` is more important on `flock()`ed files than regular files because `flock()` maintains locking semantics across process images, so (for example) clipserve may then end up holding the lock. It would be cleaner elsewhere, too, but less essential.

We do `fork()`/`exec()` on `run_clipserve()`.

cdown:
Thanks!

PR #222, opened 2024-06-06
Reload config on SIGHUP

N-R-K:
Seems to work okay from some cursory tests. But since the code wasn't written with a reload-able config in mind, there might be bugs lurking.

Keeping this a draft now, until a more careful and thorough review.

### DIFF
--- a/src/clipctl.c
+++ b/src/clipctl.c
@@ -72,7 +72,7 @@ static bool _nonnull_ should_enable(struct config *cfg, const char *mode_str) {
 
 int main(int argc, char *argv[]) {
     _drop_(config_free) struct config cfg = setup("clipctl");
-    die_on(argc != 2, "Usage: clipctl <enable|disable|toggle|status>\n");
+    die_on(argc != 2, "Usage: clipctl <enable|disable|toggle|reload|status>\n");
 
     pid_t pid = get_clipmenud_pid();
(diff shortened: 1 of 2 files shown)


cdown:
Unless there's substiantial demand I'd be inclined to avoid config reloading, or at least do it without the minefield of signal involvement.

N-R-K:
This was more of a "other deamons do it and seems like easy to add support for" rather than "i need this feature" thing. If someone really needed to have a new config, just restarting clipmenud works fine and retains previous clips. So yeah, this is probably not worthwhile.

- - -

Slightly off topic, but what exactly is `max_clips_batch` supposed to do? This is the only place where it's used, and it doesn't really make much sense to me:

```c
    if ((int)cur_clips > cfg.max_clips_batch) {
        expect(cs_trim(&cs, CS_ITER_NEWEST_FIRST, (size_t)cfg.max_clips) == 0);
    }
```

Was it meant to be `cur_clips > cfg.max_clips + cfg.max_clips_batch`?

EDIT: see https://github.com/cdown/clipmenu/pull/225

N-R-K:
> just restarting clipmenud works fine

This made me realize, you could simply re-exec instead of re-reading config and get a fully clean state. Updated the patch. Let me know if you find this approach acceptable. Otherwise I can drop the PR, as it's not too important for me.

PR #223, opened 2024-06-07
Misc fixes and improvements

N-R-K:
#### clipmenud: no need to use volatile sig_atomic_t

since there's no signal handlers involved and signals are
handled with regular code via signalfd.

#### clipmenud: don't silently ignore arguments

otherwise running something like `clipmenud --verbose` gives the
false impression of it "working".

#### config: range check parsed integers

negative values don't make sense for these config vars, so
reject them.

also reject values that are too large and would get truncated.

### DIFF
--- a/src/clipmenud.c
+++ b/src/clipmenud.c
@@ -25,7 +25,7 @@ static struct clip_store cs;
 static struct config cfg;
 static Window win;
 
-static volatile sig_atomic_t enabled = 1;
+static int enabled = 1;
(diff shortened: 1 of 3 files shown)


cdown:
These look fine, thanks. My general feeling is that if one willingly shoots themself in the foot that's fine, but I am also fine with not providing the bullets :)

And thanks for noticing we no longer need sig_atomic_t, that's from before the signalfd migration.

PR #225, opened 2024-06-08
Correct type and fix deletion threshold check

N-R-K:
The current check doesn't make much sense. With the default values, this check will always be true when there's more than 100 clips but no deletion happens until there's at least 1000 clips.

I'm assuming this was meant to be `max_clips + batch`, which would make it work similarly to the old bash script's `CM_MAX_CLIPS_THRESH`.


cdown:
Thanks!

PR #227, opened 2024-06-11
clipmenud: add option to deduplicate entries

### LINKED ISSUE #224: Deduplicate entries
Currently, duplicate snips get multiple entries clogging up the list of clips. This is especially problematic when you're using a small number of max clips (e.g 32). It'd be nice to avoid duplicate entries from being registered.

N-R-K:
uses a linear search to find the duplicate and a memmove to move it over to the end.

not perticularly efficient since each snip is 256 bytes while we're only interested in the first 8. and so iterating over it linearly like this isn't very cache friendly. the memmove worst case can also be around 250KiB with the default config.

Closes: https://github.com/cdown/clipmenu/issues/224

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -528,27 +528,52 @@ int cs_content_get(struct clip_store *cs, uint64_t hash,
 }
 
 /**
+ * Move the entry with the specified hash to the newest slot.
+ *
+ * @cs: The clip store to operate on
+ * @hash: The hash of the entry to move
+ */
+static int cs_make_newest(struct clip_store *cs, uint64_t hash) {
+    _drop_(cs_unref) struct ref_guard guard = cs_ref(cs);
+    if (guard.status < 0) {
+        return guard.status;
+    }
+
+    for (int i = 0; i < (int)cs->local_nr_snips; ++i) {
+        if (cs->snips[i].hash == hash) {
+            struct cs_snip tmp = cs->snips[i];
+            memmove(cs->snips + i, cs->snips + i + 1,
+                    (cs->local_nr_snips - (i + 1)) * sizeof(*cs->snips));
+            cs->snips[cs->local_nr_snips - 1] = tmp;
+            return 0;
+        }
+    }
+    die("unreachable");
+}
+
+/**
  * Add a new content entry to the clip store and content directory.
  *
  * @cs: The clip store to operate on
  * @content: The content to add
  * @out_hash: Output for the generated hash, or NULL
+ * @dupe_policy: Policy to use for duplicate entries
  */
-int cs_add(struct clip_store *cs, const char *content, uint64_t *out_hash) {
+int cs_add(struct clip_store *cs, const char *content, uint64_t *out_hash,
+           enum cs_dupe_policy dupe_policy) {
     uint64_t hash = djb64_hash(content);
     char line[CS_SNIP_LINE_SIZE];
     size_t nr_lines = first_line(content, line);
-
-    int ret = cs_content_add(cs, hash, content, CS_DUPE_KEEP_ALL);
-    if (ret < 0) {
-        return ret;
-    }
 
     if (out_hash) {
         *out_hash = hash;
     }
 
-    return cs_snip_add(cs, hash, line, nr_lines);
+    int ret = cs_content_add(cs, hash, content, dupe_policy);
+    if (ret == -EEXIST && dupe_policy == CS_DUPE_KEEP_LAST) {
+        return cs_make_newest(cs, hash);
+    }
+    return ret ? ret : cs_snip_add(cs, hash, line, nr_lines);
 }
 
 /**
--- a/src/store.c
+++ b/src/store.c
@@ -401,10 +401,12 @@ static int _must_use_ _nonnull_ cs_snip_add(struct clip_store *cs,
  * @cs: The clip store to operate on
  * @hash: The hash of the content to add
  * @content: The content to add to the file
+ * @dupe_policy: If set to CS_DUPE_KEEP_LAST, will return with -EEXIST when
+ * trying to insert duplicate entry.
  */
-static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
-                                               uint64_t hash,
-                                               const char *content) {
+static int _must_use_ _nonnull_
+cs_content_add(struct clip_store *cs, uint64_t hash, const char *content,
+               enum cs_dupe_policy dupe_policy) {
     bool dupe = false;
 
     char dir_path[CS_HASH_STR_MAX];
@@ -412,7 +414,7 @@ static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
 
     int ret = mkdirat(cs->content_dir_fd, dir_path, 0700);
     if (ret < 0) {
-        if (errno != EEXIST) {
+        if (errno != EEXIST || dupe_policy == CS_DUPE_KEEP_LAST) {
             return negative_errno();
         }
         dupe = true;
@@ -537,7 +539,7 @@ int cs_add(struct clip_store *cs, const char *content, uint64_t *out_hash) {
     char line[CS_SNIP_LINE_SIZE];
     size_t nr_lines = first_line(content, line);
 
-    int ret = cs_content_add(cs, hash, content);
+    int ret = cs_content_add(cs, hash, content, CS_DUPE_KEEP_ALL);
     if (ret < 0) {
         return ret;
     }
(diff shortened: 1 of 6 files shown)


N-R-K:
> the memmove worst case [...]

I was thinking, maybe it might be worth implementing tombstone now and instead of memmoving, simply marking the older entry as "dead" to be removed later.

Though I'd assume that the linear search is probably the bigger issue. Maybe maintaining a hashmap of `hash -> index`? That seems fairly complicated though given the existing clip-store design.

A simpler method might be to have a probabilistic cache of `hash -> index` instead. It'd give us an index to start the (circular) search. This way we won't have to worry about the cache being consistent. It can be wrong, but that just means we'll end up with a linear search. And when it's correct, we get to find the entry faster.

- - -

For now though, I haven't really noticed much of an issue in practice. Though I use a fairly low max clip size of 128. And both the linear search + memmove only ever happens when there's an actual duplicate, not on each insertion.

If performance becomes an issue, I can look into doing the probabilistic cache or tombstone, whichever is the bottleneck.

src/store.c
```diff
@@ -437,6 +439,23 @@ static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
...
         return 0;
+    } else if (dupe && deduplicate) {
+        _drop_(cs_unref) struct ref_guard guard = cs_ref(cs);
+        if (guard.status < 0) {
+            return guard.status;
+        }
+
+        for (int i = 0; i < (int)cs->local_nr_snips; ++i) {
+            if (cs->snips[i].hash == hash) {
+                // move cs->snips[i] to the end of the array
```
cdown: Good inline comments (as opposed to function/datastructure documentation) generally document why, not how, so not immediately seeing a reason to have this one: the code seems fairly self-explanatory.

src/store.c
```diff
@@ -437,6 +439,23 @@ static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
...
+                struct cs_snip tmp = cs->snips[i];
+                memmove(cs->snips + i, cs->snips + i + 1,
+                        (cs->local_nr_snips - (i + 1)) * sizeof(*cs->snips));
+                cs->snips[cs->local_nr_snips - 1] = tmp;
+                return 1;
+            }
+        }
+        return -ENOENT;
```
cdown: This would be a race since we already think it's a dupe, so use `expect()`.

src/store.c
```diff
@@ -546,7 +566,7 @@ int cs_add(struct clip_store *cs, const char *content, uint64_t *out_hash) {
         *out_hash = hash;
     }
 
-    return cs_snip_add(cs, hash, line, nr_lines);
+    return (ret == 1) ? 0 : cs_snip_add(cs, hash, line, nr_lines);
```
cdown: Once you use `expect()` you can avoid this and just pass through again.

src/store.c
```diff
@@ -767,8 +787,8 @@ int cs_replace(struct clip_store *cs, enum cs_iter_direction direction,
     size_t nr_lines = first_line(content, line);
     uint64_t hash = djb64_hash(content);
     cs_snip_update(snip, hash, line, nr_lines);
-    ret = cs_content_add(cs, hash, content);
-    if (ret) {
+    ret = cs_content_add(cs, hash, content, 0);
```
cdown: This is a good example of why `bool` variables are generally not a good idea: this contextless `0` now requires going to `cs_content_add` to find out what it's for.

Instead:

```
enum cs_dupe_policy { CS_DUPE_KEEP_ALL, CS_DUPE_KEEP_LAST };
```

The argument name can be `dupe_policy`.

src/store.c
```diff
@@ -437,6 +439,23 @@ static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
         }
 
         return 0;
+    } else if (dupe && deduplicate) {
```
cdown: (edit: please see below -- `cs_content_add` isn't the right place anyway)

cdown (changes requested):
Thanks!

cdown:
Also to note: I in general am a bit worried about performance here, so I should be clear I am not certain this will be merged yet. But also this should not be the regular case: we already do cheap "last dupe" checks up front.

src/store.c
```diff
@@ -437,6 +439,23 @@ static int _must_use_ _nonnull_ cs_content_add(struct clip_store *cs,
...
+        _drop_(cs_unref) struct ref_guard guard = cs_ref(cs);
+        if (guard.status < 0) {
+            return guard.status;
+        }
+
+        for (int i = 0; i < (int)cs->local_nr_snips; ++i) {
+            if (cs->snips[i].hash == hash) {
+                // move cs->snips[i] to the end of the array
+                struct cs_snip tmp = cs->snips[i];
```
cdown: This logic is in the wrong place. This function is for adding content, but this adjusts the snip cache. So this should go into a snip related function, not here.

N-R-K:
> I in general am a bit worried about performance here

I did a bit of micro benchmarking a couple days ago. It's not super scientific but the observation are as follows:

* When the duplicate is near the "top" (i.e recent) already, the operation is very fast. Roughly 200~400ns (note that the linear search currently starts at the "bottom").
* When duplicate is near the bottom, it takes considerably more time: 2~8us.
* Since adding a cache to try and speed up the linear search was relatively non invasive change, I tried it out but it barely made any difference.

These numbers are with the default of 1k max clips.

All of this seems to suggest to me that the linear search is not the bottleneck but the `memmove` is. So doing tombstone is probably more important for performance. It'd be a bit more invasive change to iteration functions and removal functions, but I think it can be done if needed.

- - -

The rest of the review all seems very reasonable. I will amend the patch and address them sometime tomorrow. Cheers.

cdown:
I think that sounds fine: my main concern was the nominal case (dupe at head), but that's already handled by is_possible_partial in a faster way, so that shouldn't present a problem. Thanks!

src/store.c
N-R-K: `cs_snip_add` resizes the store and increments `nr_snips` by +1. We don't really want that in the dedup case.

Unless by "pass through" you meant passing through the return value `ret` as is.

N-R-K:
Any blockers here?

cdown:
Sorry to be late to get back -- I don't think so :-) I have an old test suite for store.c which I haven't put in master yet and wanted to have that first, but mysteriously I couldn't find it last time I looked.

I'm away right now with pretty limited internet access, but poke me if I don't get back at the end of Jan? Thanks again for working on this

N-R-K:
> poke me if I don't get back at the end of Jan?

ping

cdown:
Thanks for your patience and work! I think we are ready for merge. I added the test suite to develop, and have rebased and added the tests for dupe policy. Want to check the tree at [0e8bc94df2c41d66ebb7c8eb6d333c8437e3e49d](https://github.com/cdown/clipmenu/commits/0e8bc94df2c41d66ebb7c8eb6d333c8437e3e49d/) to see if it looks okay to you? I merged the test update into your first commit so that bisect still works.

N-R-K:
Looks mostly fine, but shouldn't it try adding the fist duplicate in the middle or beginning? Otherwise the duplicate entry is already the newest, isn't it?

```diff
diff --git a/tests/test_store.c b/tests/test_store.c
index 9c12eaa..0d91613 100644
--- a/tests/test_store.c
+++ b/tests/test_store.c
@@ -628,10 +628,10 @@ static bool test__cs_add__dupe_keep_last_with_multiple_entries(void) {
     uint64_t hash_a, hash_dup;
     int ret = cs_add(&cs, "A", &hash_a, CS_DUPE_KEEP_ALL);
     t_assert(ret == 0);
-    ret = cs_add(&cs, "B", NULL, CS_DUPE_KEEP_ALL);
-    t_assert(ret == 0);
     ret = cs_add(&cs, "duplicate", &hash_dup, CS_DUPE_KEEP_ALL);
     t_assert(ret == 0);
+    ret = cs_add(&cs, "B", NULL, CS_DUPE_KEEP_ALL);
+    t_assert(ret == 0);
     t_assert(cs.header->nr_snips == 3);
     /* Now add a duplicate entry with KEEP_LAST which should move the duplicate
      * to the newest slot */
```

Also, I'd add an `#undef NDEBUG` in the test source since it's including user's CFLAGS/CPPFLAGS which might contain `-DNDEBUG` rendering those `assert`s as no-op.

```diff
--- a/tests/test_store.c
+++ b/tests/test_store.c
@@ -1,3 +1,4 @@
+#undef NDEBUG
 #include <assert.h>
 #include <dirent.h>
 #include <errno.h>
```

cdown:
good feedback, thank you! those are done as 51261f6. if you want, you can fetch and hard reset your branch to it, and then we can merge through this pr, or i can just manually merge -- just let me know

N-R-K:
I've separated the `NDEBUG` change into it's own commit, in order to give it a bit more context. Otherwise, everything else is kept the same.

cdown:
That's great, thank you very much! And thank you for all your help with clipmenu 7. Not long until release now :)

PR #229, opened 2024-06-23
clipdel: add -F to match literal instead of regex

N-R-K:
this adds an -F flag similar to grep to allow matching literals instead of a regex.

useful in scripting context where the script has a literal line to match against. using -F avoids having to escape troublesome regex characters.

### DIFF
--- a/src/clipdel.c
+++ b/src/clipdel.c
@@ -3,6 +3,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <string.h>
 #include <unistd.h>
 
 #include "config.h"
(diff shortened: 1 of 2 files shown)


src/clipdel.c
```diff
@@ -24,17 +25,24 @@ enum delete_mode {
...
  * be running DELETE_REAL.
  */
-static enum cs_remove_action _nonnull_
-remove_if_rgx_match(uint64_t hash _unused_, const char *line, void *private) {
+static enum cs_remove_action _nonnull_ remove_if_match(uint64_t hash _unused_,
+                                                       const char *line,
+                                                       void *private) {
     struct clipdel_state *state = private;
-    int ret = regexec(&state->rgx, line, 0, NULL, 0);
+    int ret = state->literal_match
```
cdown: patch looks good, my only thought is that the nested ternary isn't the easiest to read. maybe just do it in two if branches? then you don't need to coerce to REG_NOMATCH, too.

cdown:
Oh, and sorry, could you add a test to tests/x_integration_tests? :-)

cdown:
Looks great, thank you!

N-R-K:
I had to change the shebang to `/bin/bash` since `/usr/bin/bash` doesn't exist on my system, but then I noticed that I don't have `xsel` installed currently either. I added a test case, let's see if it passes the CI :)

cdown:
oh, that path is wrong, it should be /usr/bin/env bash. let me fix that. thanks!

PR #237, opened 2024-11-09
clipctl: Use exponential backoff for clipmenud state change detection

### LINKED ISSUE #234: clipctl disable/enable is too slow with clipmenu 7
Nowadays, because `xdotool type --delay 0 "$string"` mangles non-ASCII characters, I do this:

1. `clipctl disable`
2. Putting `$string` in the clipboard.
3. Pasting the clipboard in the target window.
4. Restoring the content of the clipboard.  (By the way, shouldn’t `clipctl enable` restore itself the content of the clipboard to whatever is first in the list shown by `clipmenu`?  Now I’m doing it by finding the most recently modified directory in `$XDG_RUNTIME_DIR/clipmenu.7.$UID` and putting the content of its file `1` in the clipboard.)
5. `clipctl enable`

But since I updated to the unreleased (I think) clipmenu 7, to take advantage of the fact that it doesn’t overwrite clips with the same first line, Steps 1 and 5 take a bit more than 100 ms each, which leads to a sensible delay when ‘typing’ in this way.

By looking at the code, I see this is due to a sleep command:
https://github.com/cdown/clipmenu/blob/1113ca8df96615aba67aa4494c7c43fb56f3dfde/src/clipctl.c#L96

I understand it is needed to ensure clipmenu is really disabled/enabled before quitting, but isn’t 100 ms too much?  Are there circumstances where it takes that long to disable itself?

As a workaround, I tried just backgrounding it in step 1 above: `clipctl disable &`, but this doesn’t leave enough time for clipmenud to disable itself and the `$string` is recorded by clipmenud.

Now I’m using another workaround: when launching clipmenud in xinitrc, I cache its PID in `$XDG_RUNTIME_DIR/clipmenu.7.$UID/pid` and in step 1 above, I just do `kill -USR1 "$(<"$XDG_RUNTIME_DIR/clipmenu.7.$UID/pid")"` as the old Bash clipctl did (without the caching of the PID).  With that, I didn’t have a leak of `$string` in any of my tests.

So I’m wondering if it would not be possible to check every 1 ms or so rather than every 100 ms, or first every 1 ms and then every 100 ms for the exceptionally slow cases?

cdown:
This was previously hardcoded to 100ms, which is way too high.

Fixes #234.

### DIFF
--- a/src/clipctl.c
+++ b/src/clipctl.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/types.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "config.h"
@@ -66,10 +67,6 @@ static bool _nonnull_ should_enable(struct config *cfg, const char *mode_str) {
     die("Unknown command: %s\n", mode_str);
 }
 
-/* The number of times to check if state was updated before giving up */
-#define MAX_STATE_RETRIES 20
-#define USEC_IN_MS 1000
-
 int main(int argc, char *argv[]) {
     _drop_(config_free) struct config cfg = setup("clipctl");
     die_on(argc != 2, "Usage: clipctl <enable|disable|toggle|status>\n");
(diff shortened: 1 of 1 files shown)


src/clipctl.c
```diff
@@ -89,11 +89,30 @@ int main(int argc, char *argv[]) {
     expect(kill(pid, want_enable ? SIGUSR2 : SIGUSR1) == 0);
     dbg("Sent signal to pid %d\n", pid);
 
+    unsigned int delay_ms = 1;
+    unsigned int total_wait_ms = 0;
+    const unsigned int max_wait_ms = 1000;
+
     for (size_t retries = 0; retries < MAX_STATE_RETRIES; ++retries) {
```
N-R-K: `MAX_STATE_RETRIES` seems redundant now. Especially since (if my math is right) it will never be reached due to `total_wait_ms` reaching 1023 after 10 iterations.

I suggest gutting it and changing the loop condition like so:

```suggestion
    while (total_wait_ms < max_wait_ms) {
```

N-R-K (changes requested):

src/clipctl.c
cdown: oh yes, that was silly to leave in. thanks

PR #239, opened 2024-11-15
store: use fixed length hex hashes for dir names

N-R-K:
mostly cosmetics. having fixed width dirnames results in a nicer listing of the directory. it also matches usual expectation of hashes being stringified in hex format.

image

### DIFF
--- a/src/clipserve.c
+++ b/src/clipserve.c
@@ -149,7 +149,7 @@ static void _nonnull_ serve_clipboard(uint64_t hash,
 
                 _drop_(XFree) char *window_title =
                     get_window_title(dpy, req->requestor);
-                dbg("Servicing request to window '%s' (0x%lx) for clip %" PRIu64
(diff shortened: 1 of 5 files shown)


cdown:
I'm not sure I understand the justification, could you go over it more?

N-R-K:
With the current naming, directories would look something like this:

image

With fixed-width hex naming, they all have the same width and appear nicer when listing (or when viewed thru a file manager etc).

It's entirely a cosmetic/aesthetic thing. Figured I'd raise it before there's a stable version released in case you like it, otherwise I can just keep the patch locally if you don't.

cdown:
I was on the fence because while it does make hashes more uniform, it also adds (ok, not much) more work and code. I do think it looks better and is also visually more obvious when seeing clip hashes in CM_DEBUG, though, so let's do it. Thanks!

PR #245, opened 2025-03-12
clipdel: add options to delete number of entries

N-R-K:
Adds an `-n` flag which allows deleting certain number of entries from newest (default) or oldest (`-N`).

Allows deleting the current clip with `clipdel -n 1` as requested in https://github.com/cdown/clipmenu/issues/57#issuecomment-366728601 and https://github.com/cdown/clipmenu/issues/98.

Caveat: `-n 1` selects the newest/oldest based on time of entry. But the currently serving clip maybe different if another entry was selected thru `clipmenu`.

Opinion on adding a config option to pull an entry to the top when selecting it to be served via clipmenu? I'd like this personally, since I find it more natural to see the current serving entry on the top when opening clipmenu.

### DIFF
--- a/man/clipdel.1
+++ b/man/clipdel.1
@@ -17,6 +17,17 @@ Real deletion mode. Matching clipboard entries will be removed.
 .B \-F
 Perform a literal (fixed-string) match instead of interpreting the pattern as a regular expression.
 .TP
+.B \-N
+Interpret PATTERN as the amount of entries to delete starting from the oldest.
+E.g
+.I "\-N 1"
+deletes the oldest entry.
+.TP
+.B \-n
+Same as
+.B \-N
+but starting from newest.
+.TP
 .B \-v
 Invert the matching condition; entries that do not match the given pattern are selected for deletion.
 .TP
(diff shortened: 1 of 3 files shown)


snprajwal (commented):
Thank you for this! This will allow us to enable better functionality with passwordstore :)

cdown:
(concept looks good, may be a bit to review. thanks for your patience)

cdown:
This looks great, thank you. I guess the only thing remaining is an integration test?

N-R-K:
> I guess the only thing remaining is an integration test?

I thought I had added them already but apparently not. Added them now. But it seems to be failing the incr test, which seems unrelated (?)

cdown:
weird that it was flaky on this and the other pr. let me take a look at what's going on there

(and github, why on earth can't you retry a pr workflow...)

cdown:
That actually looks like some real bug rather than a failure, it's failing dupe detection somehow, or some other clip is getting in. I will add more instrumentation to the integration tests to print out the storage so that if it happens again we find out what actually happened.

Anyway, it passed when run again on CI and locally, so merging. Longer term hopefully those will allow us to debug the flakiness or bug. Thank you!

cdown:
087cf07022eca27eaa5b69d53fc3809ae7afe801 should hopefully work out what's going on there.

PR #246, opened 2025-03-12
config: add touch_on_select

N-R-K:
this makes an entry bump to the newest slot when serving (e.g selecting via `clipmenu`).

### DIFF
--- a/man/clipmenu.conf.5
+++ b/man/clipmenu.conf.5
@@ -47,6 +47,12 @@ include rofi's dmenu mode or a custom command. Default: "dmenu".
 When enabled, extra command-line arguments passed to clipmenu are forwarded to
 the launcher. Default: 1.
 .TP
+.B touch_on_select
+When an entry is selected via
+.BR clipmenu (1)
+it will also be moved up to the newest slot.
+Default: 0.
+.TP
 .B cm_dir
 Overrides the default directory for the clip store. This is by default at a
 subdirectory inside XDG_RUNTIME_DIR, TMPDIR, or if both are unset, inside /tmp.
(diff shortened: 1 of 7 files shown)


man/clipmenu.conf.5
```diff
@@ -47,6 +47,10 @@ include rofi's dmenu mode or a custom command. Default: "dmenu".
 When enabled, extra command-line arguments passed to clipmenu are forwarded to
 the launcher. Default: 1.
 .TP
+.B touch_on_serve
```
N-R-K: Open to better name suggestions :)

cdown:
Hmm, clipserve is somewhat internal and can also be called when starting clipboard ownership. What are your thoughts on doing it from clipmenu when selecting the clip?

N-R-K:
> Hmm, clipserve is somewhat internal and can also be called when starting clipboard ownership.

In those cases, isn't the entry already on top and thus make_newest() would be a no-op?

cdown:
I was going to say we can end up racing and bumping something from another selection, but I suppose that's no worse than the current situation since post-fork execution is unbounded, so I think you're right that it doesn't matter.

Maybe since "serve" is a bit internal, it should be "touch on select"? What do you think?

N-R-K:
> Maybe since "serve" is a bit internal, it should be "touch on select"?

Sounds fine to me.

> I was going to say we can end up racing and bumping something from another selection, but I suppose that's no worse than the current situation since post-fork execution is unbounded, so I think you're right that it doesn't matter.

Since it's called `touch_on_select` now, I guess it makes more sense to put it in `clipmenu.c` now, even if the outcome doesn't really change.

While doing that, I noticed it was using `WEXITSTATUS` without checking `WIFEXITED` first, so fixed that. Also added `EINTR` handling too.

N-R-K:
The CI failure seems to be unrelated? It [passed earlier](https://github.com/N-R-K/clipmenu/actions/runs/14030873676/job/39278021677) on my repo.

cdown:
reran it and everything looks great, thank you!

PR #250, opened 2025-03-27
config: more friendly error message on mkdir failure

N-R-K:
### config: more friendly error message on mkdir failure

mkdir may fail in case user sets cm_dir to something
non-existent expecting clipmenu to create the parent dir as
well. give a nicer error message on such failure.

### ~~config, store: set the sticky bit on directories~~

~~clipmenu uses XDG_RUNTIME_DIR if set, which according to the
spec [^0] is subject to cleanup every 6 hours, unless the sticky
bit is set. we almost certainly do not want this, so set the
sticky bit on the directories we create.~~

[^0]: https://specifications.freedesktop.org/basedir-spec/latest/

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -44,7 +44,8 @@ char *get_cache_dir(struct config *cfg) {
     // In case config changed, do the write anyway
     snprintf_safe(cache_dir, PATH_MAX, "%s/clipmenu.%d.%ld", cfg->runtime_dir,
                   CLIPMENU_VERSION, (long)getuid());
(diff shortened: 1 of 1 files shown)


N-R-K:
Actually, the spec specifically says to put the sticky bit on file, not directory. And I don't think anything is doing this cleanup in practice: a lot of other stuff seems to neither set the sticky bit nor update the access time. I'll remove this commit (https://github.com/cdown/clipmenu/pull/250/commits/d21f3d95a0e5d9be0029bbdb36ea2c57a9d90872).

The other commit about nicer error message should still be fine to apply.

cdown:
Thank you!

PR #251, opened 2025-04-04
test_store: unlink test files during teardown

### LINKED ISSUE #249: Don't leave behind test files and dirs
The current `make tests` seems to leave some files at `/dev/shm/*` laying around. This causes problems when, e.g the tests are run through some package manager and thus created with a different user and vice versa:

```console
[clipmenu develop]~> ls -lAh /dev/shm   # files left behind from earlier install thru portage
total 4.0K
drwx------ 5 portage portage  100 Mar 24 13:58 clip_store_content_dir_test
-rw------- 1 portage portage 257K Mar 24 13:58 clip_store_snip_test
[clipmenu develop]~> make -s tests
test__cs_init:
test_store: tests/test_store.c:51: create_test_snip_fd: Assertion `snip_fd >= 0' failed.
make: *** [Makefile:108: tests] Aborted
```

Expectation: `make tests` should cleanup after itself properly and not leave any files/dirs laying around.

N-R-K:
…rather than doing it at creation.

Fixes: https://github.com/cdown/clipmenu/issues/249

### DIFF
--- a/tests/test_store.c
+++ b/tests/test_store.c
@@ -146,7 +146,9 @@ static void drop_teardown_test(struct clip_store *cs) {
     int ret = cs_destroy(cs);
     assert(ret == 0);
     close(snip_fd);
+    shm_unlink(TEST_SNIP_FILE);
     close(content_dir_fd);
+    remove_test_content_dir(TEST_CONTENT_DIR);
 }
 
 static void add_ten_snips(struct clip_store *cs) {


cdown:
I think maybe both are the right solution? I did it this way so that even if we have some existing state it's wiped.

N-R-K:
> I think maybe both are the right solution? I did it this way so that even if we have some existing state it's wiped.

Sounds good. Added them back.

cdown:
capital, thank you!

PR #254, opened 2025-05-28
clipmenud: Prevent race in disable/enable by checking X timestamps

### LINKED ISSUE #217: Check serial time as part of disable rejection
It's possible (but highly unlikely) right now we may race on disable/enable if there was a lot of clipboard spam. It would be better to record the current X server time at the point we last disabled and last enabled, and only accept selections from between then.

cdown:
When clipmenud is rapidly disabled and reenabled, stale X11 selection events queued while disabled could be processed after reenabling, causing entries we should have thrown away to be stored.

This is theoretical, there are no user reports of issues.

Fixes #217.

### DIFF
--- a/src/clipmenud.c
+++ b/src/clipmenud.c
@@ -29,10 +29,14 @@ static Window win;
 static int enabled = 1;
 static int sig_fd;
 
+static Atom timestamp_atom;
 static Atom incr_atom;
 static struct incr_transfer *it_list;
 
 static struct cm_selections sels[CM_SEL_MAX];
+
+static Time last_disable_time = 0;
+static Time last_enable_time = 0;
 
 enum clip_text_source {
     CLIP_TEXT_SOURCE_X,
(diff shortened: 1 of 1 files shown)


src/clipmenud.c
```diff
@@ -60,6 +64,25 @@ static void free_clip_text(struct clip_text *ct) {
     ct->source = CLIP_TEXT_SOURCE_INVALID;
 }
 
+/**
+ * Get the current X server time by triggering a PropertyNotify.
+ */
+static Time get_current_server_time(void) {
```
cdown, line 70: @N-R-K do you know a better way to do this? i wondered if there was some extension that can help here, but it seems not?

N-R-K: > i wondered if there was some extension that can help here, but it seems not?

@cdown Well, yes, I *do* happen to have a bit of [esoteric knowledge](https://nrk.neocities.org/articles/x11-timeout-with-xsyncalarm) in this matter. Roughly, you'd need to do the following:

1. `XSyncQueryExtension()` + `XSyncInitialize()` at startup for initialization.
2. Then use `XSyncListSystemCounters()` to get a list of available counters.
3. Iterate over them to find "SERVERTIME".
4. Now you should be able to use `XSyncQueryCounter()` to query the server time with the counter id found via step 3.

But I'm not entirely sure if this is worth the trouble. Can't client set arbitrary timestamps anyways? Though I guess from a security perspective it makes sense to completely eliminate any chances of storing a sensitive clip that should've been discarded.

cdown: Thank you! I guessed you might have some knowledge here :-)

Okay, I guess we'll stick with the PropertyNotify method for now since we don't need the XSync guarantees (and it's less code).

PR #260, opened 2026-02-12
clipdel: add -q to suppress output

temporecras:
While working on an script involving clipmenu I ran into it printing all the pastes to be deleted to stdout. This was a major inconvenience as I was trying to delete sensitive information following a list of regexs.

This PR adds a -q (quiet) flag to allow the output to be suppressed natively without the need for workarounds (as redirecting output to /dev/null).

Changes:
added -q flag to clipdel.c
updated clipdel.1 man page
Added integration tests in tests/x_integration_tests to verify the quiet behavior

### DIFF
--- a/man/clipdel.1
+++ b/man/clipdel.1
@@ -31,6 +31,9 @@ but starting from newest.
 .B \-v
 Invert the matching condition; entries that do not match the given pattern are selected for deletion.
 .TP
+.B \-q
+Quiet mode. Do not print matching entries to standard output.
+.TP
 .B \-h, \--help
 Display the help message (invokes the manual page).
 .SH CONFIGURATION
--- a/src/clipdel.c
+++ b/src/clipdel.c
@@ -36,6 +36,7 @@ struct clipdel_state {
     enum delete_mode mode;
     enum match_type type;
     bool invert_match;
+    bool quiet;
     union {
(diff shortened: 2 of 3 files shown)


N-R-K:
> as redirecting output to /dev/null

This seems like a valid and universal solution to me. What's the problem with this?

temporecras:
Redirection works, and that's a fair point, but native support for quietness is common in similar utilities for better ergonomics. It keeps scripts cleaner and I believe it's a low-maintenance addition that makes the tool more intuitive.

temporecras:
Hi! Just checking on this PR as it's been a while. Happy to make any change if needed.

cdown:
Apologies for not getting back sooner.

My opinion is basically the same as @N-R-K I guess. I don't feel strongly about it either way, but I also think there's the question to ask of:

1. If this is added to clipdelmenu, are there other tools it may need adding to now or in the future for consistency?
2. If so, how invasive will that be?

This is a pretty weak argument since I doubt clipmenu as a suite of tools will grow that much, or the functionality either, but using `>/dev/null` avoids having to worry about it in the first place.

So I guess I weakly prefer to leave it as is.

temporecras:
I get your point and understand it makes sense to leave it as is. Thanks for your feedback!

